Issue #54, opened 2015-01-27
Merging coverage of different tests using the same executable


mgehre:
Hi,

thanks again for fixing the shared library coverage. My current issue lies with the merging of coverage from multiple runs of the same executable.
If I run
```
kcov /tmp/kcov app -opt1
kcov /tmp/kcov app -opt2
```
then /tmp/kcov holds only the results of the last invocation of "app".

If I use different directories for each invocation and then use kcov-merge, I seem not to be able to look at individual source file coverage anymore.

Is there a way to aggregate coverage from multiple runs of the same executable (with different options?)

Best wishes,
Matthias

SimonKagstrom:
That's actually something which should just work "out-of-the-box", and I also have regression tests for it in the test suite, which passes.

What kind of application are you running (Python, Bash or a compiled program)?


Does it work with a trivial test, such as a shell script which can be invoked with different parameters:

```sh
#!/bin/bash

if [ $1 -eq 1 ] ; then
   echo "first"
else
   echo "second"
fi
```

kcov-merge only outputs a summary, so individual sources are not present there. Something similar would be possible to add to kcov itself, but I haven't seen it as very important so far.

SimonKagstrom:
Is this in a shared library by the way?

In that case, it might be caused by the library being loaded at a different base address the next time. If this is the case, the bug is caused by kcov storing collected data with actual addresses (which are not meaningful for libraries).

mgehre:
Yes, its in a shared library. I didn't check if it merges the actual exectuable coverage correctly, I was just looking at the library.

SimonKagstrom:
OK. I didn't realize that this was a problem until now. I'll add a regression test and think of a solution.

SimonKagstrom:
Fixed (but lightly tested!) via 96c6eeb.

The implementation remembers the address ranges when shared libraries are added, and stores rangeIndex,offset pairs in the database.


Closing, but feel free to reopen if it turns out to be buggy :-)

PR #77, opened 2015-03-14
Add initial support for position-independent executables.

lifthrasiir:
What an appropriate PR for [pi(e) day](https://en.wikipedia.org/wiki/Pi_Day), heh. :) Anyway, this PR adds two features that enable the support for position-independent executables:

* Disables the address randomization after fork. While this does not directly enable PIE support (AFAIK, x86_64 Linux always relocates the PIE at 0x555555554000) this should help any potential issues from ASLR.
* Reads the "base address" (or more accurately, relocation) of PIE from `solib_wrapper`. This has some design implications, and I went to the easiest way possible: DWARF parsing is deferred until `solib_wrapper` returns. I think we already have to wait for that, so this should not be a big problem.

Again with the prior PR, I haven't written a proper test case for this. (I tried to run test cases, but I'm blocked on pybot downloads...) Any executable compiled with `gcc -pie -fpie -fPIE` should work.

### DIFF
--- a/src/engines/ptrace.cc
+++ b/src/engines/ptrace.cc
@@ -7,6 +7,7 @@
 #include <phdr_data.h>
 
 #include <unistd.h>
+#include <sys/personality.h>
 #include <sys/ptrace.h>
 #include <sys/user.h>
 #include <sys/types.h>
@@ -487,7 +488,20 @@ private:
 
 		/* Executable exists, try to launch it */
 		if ((child = fork()) == 0) {
+			int persona;
 			int res;
+
+			/* Avoid address randomization */
+			persona = personality(0xffffffff);
+			if (persona < 0) {
+				perror("Can't get personality");
+				return false;
+			}
+			persona |= 0x0040000; /* ADDR_NO_RANDOMIZE */
+			if (personality(persona) < 0) {
+				perror("Can't set personality");
+				return false;
+			}
 
 			/* And launch the process */
 			res = ptrace(PTRACE_TRACEME, 0, 0, 0);
--- a/src/engines/script-engine-base.hh
+++ b/src/engines/script-engine-base.hh
@@ -50,6 +50,11 @@ public:
 		return true;
 	}
 
+	bool setMainFileRelocation(unsigned long relocation)
+	{
+		return true;
+	}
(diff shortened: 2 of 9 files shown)


SimonKagstrom:
Haha! Did you wait with the submission until pi-day or was it just lucky coincidence?

Anyway, very nice work and thanks for the submission! I can add test cases to cover this, it's not difficult to do. As for pybot, I install it with the "pip" tool, which seems to work in the Travis builds at least (see '''travis/Makefile''' for how it's setup in Ubuntu).

lifthrasiir:
@SimonKagstrom Just a lucky coincidence. ;) As the above issue linked suggests, I was trying to revive kcov for Rust programs (which are now compiled as a PIE by default, and non-PIE codegen needs some non-trivial work). Turned out that this too was a non-trivial task though.

I've finally figured out what was wrong with my local pip (`easy_install -U pip` was helpful). I've added basic tests for PIE support as #78. How do you test locally by the way? I haven't found any shortcut and had to resort with `make -B -f travis/Makefile build && cd build-tests && WORKSPACE=`pwd`/.. pybot --noncritical not_ready ../tests/robot-framework/kcov-tests.txt` (oops).

SimonKagstrom:
Locally I do more or less what you did. My build directories for the tests and kcov itself are laid out the same way as for travis (or really, the other way around), so I just do

```
WORKSPACE=`pwd`/.. pybot --noncritical not_ready ../tests/robot-framework/kcov-tests.txt
```

in the build-tests directory.


Anyway, I added a couple of more tests for PIE executables.

One thing which doesn't work for these is coverage accumulation, i.e., data from multiple runs. I added the pie-accumulate test (failing) which checks for this. It's not an error in your commit, but rather a limitation in the kcov implementation which stores raw addresses and "re-runs" these when the binary is started again. Shared libraries used to have the same bug (Issue #54), which was solved through storing an offset instead of the absolute address.

PIE executables should be possible to handle in the same way. Another possibility would be to use the same method as in Python/Bash (where a file/line hash is stored instead), but without being careful, it would mean missing information about partial hits.

Good to hear that kcov gets used by the way :-)

PR #86, opened 2015-04-25
Fixed build fail on python3 default machines

zyphrus:
kcov was failing on my Arch Linux machine, which defaults `/usr/bin/env python` to python3.

This makes it more explicit that python2 is needed


SimonKagstrom:
Merged, thanks!

I guess bin-to-source should be python3-clean, but I'll look at that some day.

PR #87, opened 2015-04-25
Add coverage writer in merge mode.

meqif:
This pull request adds support for sending the merged coverage to coveralls.io in merge mode.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -164,6 +164,7 @@ static int runMergeMode()
 			base, base + "/kcov-merged", "[merged]", true);
 	IWriter &mergeCoberturaWriter = createCoberturaWriter(mergeParser, mergeReporter,
 			base + "kcov-merged/cobertura.xml");
+	IWriter &mergeCoverallsWriter = createCoverallsWriter(mergeParser, mergeReporter);
 	(void)mkdir(fmt("%s/kcov-merged", base.c_str()).c_str(), 0755);
 
 	reporter.registerListener(mergeParser);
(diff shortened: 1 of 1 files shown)


SimonKagstrom:
Merged, thanks! I can think of one improvement though: a command-line option to select either merged or "normal" coveralls pushes.

meqif:
Thank you for merging!

I'm not entirely sure what you mean. By "normal" do you mean the output of running kcov against more than one executable with the same output directory, i.e., `kcov ./tmp/kcov ./first_executable && kcov ./tmp/kcov ./second_executable`?

I might be recalling it incorrectly, but coveralls seemed to barf when I used `--coveralls-id=…` on both runs, and to receive only the coverage of the last executable when I used the option on the last run. This made me suspect there might be an issue on coveralls's side with pushes with more than one entry in the "root". Perhaps I was doing something wrong. In any case, using kcov's merge mode and pushing the merged output proved to be a simple solution.

SimonKagstrom:
Well, with "normal" I mean that only the coverage of '''second_executable''' would be shown in your example.

But the more I think about it, perhaps it's better to just default to always reporting the merged results if there are more than one executable in the coverage report. Basically so that only one coverallsWriter is registered in each run - the previous one if there is only one executable, and your merged one when there is merged data available.

Right now both should be called, and probably coveralls will be slightly confused by that (the order is of them uploading data is not set in stone either).

SimonKagstrom:
My mistake: I didn't actually look at the code until now. So your patch is only used when running in merge-mode, which makes sense.

The change I described above is something separate, but which I think could make sense as well. But no complaints about your improvement then!

meqif:
Yeah, I was a little confused by your previous comments, since my changes are quite small and confined to `runMergeMode`. Sorry for not describing them better before.

The change you described makes sense, but I'm not comfortable enough with kcov's codebase to implement it.

PR #95, opened 2015-05-13
python-helper.py explict python version

zyphrus:
Noticed `python-helper.py` did not have an explicit version


SimonKagstrom:
The patch looks good, but is it really needed? I think I've tried running with both Python 2 and Python 3, and it's at least the intention that it should work with both.

Of course, I might have made some mistake at some point - I have Python 2 default and might not have noticed regressions :-)

PR #108, opened 2015-07-29
Describe the implemented usage of -p / --pid option, not the intended one

### LINKED ISSUE #107: When -p / --pid option is used, the executable should not have to be passed on the command line
Hi,

I'm using a build of kcov from a git snapshot at revision 157d013. The man page states the executable argument is optional if kcov is launched with `--pid`. However, if the executable isn't specified, kcov terminates with an error.

1. For instance, launch the snmpd daemon

        $ /usr/sbin/snmpd

2. When trying to attach it without giving an executable name, an error occurs

        $ ps -ef | grep snmpd
        root      8307      1  0 16:03 ?        00:00:00 /usr/sbin/snmpd
        $ kcov --pid=8307 /tmp/kcov_results
        kcov: error: Can't find or open 

3. If the executable name is given, this error doesn't show up

        $ kcov --pid=8307 /tmp/kcov_results /usr/sbin/snmpd

Best regards,

VincentGibert:
Fixes #107

### DIFF
--- a/doc/kcov.1
+++ b/doc/kcov.1
@@ -35,7 +35,7 @@ compiler directives, and continuously produces output from long-running applicat
 .SH OPTIONS
 .TP
 \fB\-p\fP, \fB\-\-pid\fP=\fIPID\fP
-Trace PID instead of executing executable, executable is optional in this case.
+Trace PID instead of executing executable.
 .TP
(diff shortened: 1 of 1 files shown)


SimonKagstrom:
Merged, thanks!

PR #111, opened 2015-08-07
typofix - https://github.com/vlajos/misspell_fixer

vlajos:


SimonKagstrom:
Merged, thanks for your patch and tool!

PR #112, opened 2015-08-12
Trivial fixes

VincentGibert:


SimonKagstrom:
Thanks! However, the --help message fix should probably be made as

```
 kcov --pid=1000 /tmp/kcov ./frodo
```

so that it's the same way as for the other examples. You example will certainly work, but the real bug is that I must have forgotten a space there :-)

VincentGibert:
Oops, fixed.

SimonKagstrom:
Merged, thanks!

Issue #116, opened 2015-10-28
Script coverage stops at << in bash script


philwhineray:
Below are a couple of examples, in both cases the coverage stops at the line `<<` for me with version 29, also with git master (do you have a preference for bug reports?).

~~~~
$ bash --version
GNU bash, version 4.3.30(1)-release (x86_64-pc-linux-gnu)
~~~~

Arithmetic:

~~~~
#!/bin/bash

echo "Start"

SUB=30
max=$[ 1 << (32 - SUB) ]

echo "$max"

echo "Not covered"
echo "Not covered"
echo "Not covered"
~~~~

Heredoc:

~~~~
#!/bin/bash

echo "Start"

cat <<-!
	Middle
!

echo "Not covered"
echo "Not covered"
echo "Not covered"
~~~~

I have seen other heredoc examples that work, so I guess it may be the use of '!' as the EOF, or the '-' that are causing that problem in that script

SimonKagstrom:
Issues are perfect for bug reports, so that's fine!

The heredoc problem should be found somewhere around here:

  https://github.com/SimonKagstrom/kcov/blob/master/src/engines/bash-engine.cc#L571

I'm not sure what causes it though. As for the arithmetic problem, I think it's because I don't check for $[, and that you have a shift-left operator in your example. That should be around here:

  https://github.com/SimonKagstrom/kcov/blob/master/src/engines/bash-engine.cc#L588


Anyway, I'm not proud of the bash scanning code. It should really be replaced by a proper parser, but I don't know bash well enough to write one myself, and haven't found a good and simple example to copy from.

SimonKagstrom:
Fixed with fff5ad9 and merge #117 .

PR #117, opened 2015-10-28
More bash fixes

philwhineray:
Thanks for looking at #116 so quickly.

I solved the unhandled heredoc syntax problems using the pointers you provided. Hopefully the first commit is good enough you can just take it.

I also hit a problem with a string containing a #, which is the second commit. It is not ideal because the string could both start and end before, so be a legitimate comment... the best I can say is it works for my use case.

### DIFF
--- a/src/engines/bash-engine.cc
+++ b/src/engines/bash-engine.cc
@@ -578,6 +578,18 @@ private:
 						}
 					}
 
+					if (heredocMarker[0] == '-') {
+						// '-' marks tab-suppression in heredoc
+						heredocMarker = heredocMarker.substr(1);
+					}
+
+					if (heredocMarker.length() > 2
+					&& (heredocMarker[0] == '"' || heredocMarker[0] == '\'')
+					&&  heredocMarker[0] == heredocMarker[heredocMarker.length()-1]) {
+						// remove enclosing in quotes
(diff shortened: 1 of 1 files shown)


SimonKagstrom:
Thanks, merged!

It would be good to add a regression test for the #-fix as well. Do you have some short example with that?

philwhineray:
Well, I just cut and pasted straight from the script I had, then
played around until I found the problem, so you might want to tidy it
up for regression purposes, but this definitely showed the problem:

~~~~
#!/bin/bash

work_runtime_error=0

runtime_error() {
        local type="ERROR" id=

        case "${1}" in
                error)
                        type="ERROR  "
                        work_runtime_error=$[work_runtime_error + 1]
                        echo beforeid
                        id="# ${work_runtime_error}."
                        echo afterid
                        ;;

                warn)
                        type="WARNING"
                        id="This might or might not affect the
operation of your firewall."
                        id="# ${work_runtime_error}."
                        ;;
                *)
                        work_runtime_error=$[work_runtime_error + 1]
        esac
        shift
        echo $type
        echo $id
}

echo Start
runtime_error error
echo Finish
~~~~

SimonKagstrom:
Thanks. Commit 63b6cf9 adds a regression test for this issue as well!

PR #132, opened 2016-01-31
fix python version in travis makefile

debris:
without python2, build fails here:

https://github.com/SimonKagstrom/kcov/blob/6ac673df3e141b28971f506d476de5b1b63ed229/src/bin-to-c-source.py#L1

```
/usr/bin/env: python2: No such file or directory
make[2]: *** [src/library.cc] Error 127
make[2]: *** Waiting for unfinished jobs....
[ 16%] /usr/bin/env: python2: No such file or directory
make[2]: *** [src/bash-redirector-library.cc] Error 127
```

### DIFF
--- a/travis/Makefile
+++ b/travis/Makefile
@@ -1,7 +1,7 @@
 #!/usr/bin/make
 
 chroot=/tmp/32-bit-chroot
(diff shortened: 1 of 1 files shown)


SimonKagstrom:
Thanks!

I have other travis build issues right now (for the kcov build), which I can't understand, but that's another problem :-)

PR #136, opened 2016-04-16
Exit with status 0 when printing version.

pietro:
Printing the version string is very useful for build scripts, specially on continuous integration set ups. Exiting with non-zero status mean an error happened which is not the case.


SimonKagstrom:
Yep, that makes sense.

Merged, thanks!

PR #146, opened 2016-05-25
support setting strip path mamuly

BusyJay:
Currently when uploading coverage data to coveralls.io, common path will be tripped away, which may lead to source not found error when you want to check the coverage data of a single file. For example, if using following command to collect coverage:

```
kcov --include-patten projectname/src --coveralls-id ${TRAVIS_JOB_ID} target/kcov path/to/executable
```

file *projectname/src/main.c* will be stripped as *main.c*, but coveralls.io requires *src/main.c* to be able to fetch source code from github.


SimonKagstrom:
Merged, thanks!

PR #169, opened 2017-01-05
Adds JSON writer

brycefisher:
@SimonKagstrom here's a super basic version of the json writer. It output the file as pure json.

TODO:

 * [x] Fix overall project line counts
 * [x] Add overal percentage covered
 * [x] Add per file summaries
 * [x] Add documentation

### DIFF
--- a/src/writers/json-writer.cc
+++ b/src/writers/json-writer.cc
@@ -1,0 +1,138 @@
+namespace std { class type_info; }
+
+#include <reporter.hh>
+#include <file-parser.hh>
+#include <configuration.hh>
+#include <writer.hh>
+#include <utils.hh>
+
+#include <string>
+#include <list>
+#include <unordered_map>
+#include <iostream>
+#include <fstream>
+
+#include "writer-base.hh"
+
+using namespace kcov;
+
+class JsonWriter : public WriterBase
+{
+public:
+	JsonWriter(IFileParser &parser, IReporter &reporter,
+			const std::string &outFile) :
+		WriterBase(parser, reporter),
+		m_outFile(outFile)
+	{
+	}
+
+	void onStartup()
+	{
+	}
+
+	void onStop()
+	{
+	}
+
+	void write()
+	{
+		std::ofstream out(m_outFile);
+
+		// Output directory not writable?
+		if (!out.is_open())
+			return;
+
+		out << "{\n";
+		out << "  \"files\": [\n";
+
+		IConfiguration &conf = IConfiguration::getInstance();
+		unsigned int nTotalExecutedLines = 0;
+		unsigned int nTotalCodeLines = 0;
+		double percentCovered = 0.0;
+		bool firstFile = true;
+
+		for (FileMap_t::const_iterator it = m_files.begin();
+				it != m_files.end();
+				++it) {
+			File *file = it->second;
+			unsigned int nExecutedLines = 0;
+			unsigned int nCodeLines = 0;
+
+			for (unsigned int n = 1; n < file->m_lastLineNr; n++) {
+				IReporter::LineExecutionCount cnt = m_reporter.getLineExecutionCount(file->m_name, n);
+				if (m_reporter.lineIsCode(file->m_name, n)) {
+					nExecutedLines += !!cnt.m_hits;
+					nCodeLines++;
+					nTotalExecutedLines += !!cnt.m_hits;
+					nTotalCodeLines++;
+				}
+			}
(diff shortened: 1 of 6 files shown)


src/writers/json-writer.cc
```diff
@@ -0,0 +1,101 @@
...
+		unsigned int nCodeLines = 0;
+
+		setupCommonPaths();
+
+		for (FileMap_t::const_iterator it = m_files.begin();
+				it != m_files.end();
+				++it) {
+			File *file = it->second;
+			nCodeLines += file->m_codeLines;
+			nExecutedLines += file->m_executedLines;
```
brycefisher: On my machine in my rust project, `file->m_executedLines` and `file->m_codeLines` seem to always be zero. However, the values are non-zero for the HTML writer. What am I missing here?

src/writers/json-writer.cc
```diff
@@ -0,0 +1,101 @@
...
+
+		// Output directory not writable?
+		if (!out.is_open())
+			return;
+
+		IConfiguration &conf = IConfiguration::getInstance();
+		unsigned int nExecutedLines = 0;
+		unsigned int nCodeLines = 0;
+
+		setupCommonPaths();
```
brycefisher: I'm still not entirely sure what this bit is doing. Copied it over from the HTML writer.

src/writers/json-writer.cc
```diff
@@ -0,0 +1,101 @@
...
+		WriterBase(parser, reporter),
+		m_outFile(outFile)
+	{
+	}
+
+	void onStartup()
+	{
+	}
+
+	void onStop()
```
brycefisher: Do I actually need the `onStop()` and `onStartup()`?

SimonKagstrom: They need to be defined, but I don't think you need to do anything there.

src/writers/json-writer.cc
SimonKagstrom: I don't think setupCommonPaths need to be there. It's used to generate the [...] stuff in HTML reports (i.e., the parts of the paths which are common to all files in the report).

src/writers/json-writer.cc
SimonKagstrom: It's probably easier to look in e.g., cobertura-writer.cc, where these are setup. They are actually calculated in the writers.

However, this should probably be moved to common code, but I'll add a separate bug for that.

brycefisher: So, when I looked at the sonarqube and cobertura writers, the lines were counted by iterating over every line in the file and incrementing a counter.

I copied this out of the HTML writer since it doesn't _appear_ to loop over line, but merely accesses these same properties and sums them from each file (which is what i'd like to do). 

I can definitely try to just iterate over the lines in each file anyway, since I want to do that for the file by file synopsis I plan to add this report; it just felt a bit weird that this didn't work and I couldn't immediately see why.

SimonKagstrom: Now that I think of it, I think it might be that the File object is used by all writers, so technically, only one need to update the stats. Of course, that shouldn't be relied on, so it should really be refactored to be handled outside. (Sorry about that!)

Anyway, I think the best way is to get the json writer working, and then refactor them all afterwards. There are probably other things which should be changed as well, now that there will be quite a few different writers.

brycefisher:
@SimonKagstrom okay, I think is more or less ready to go. Let me know what feeback you have!

SimonKagstrom:
I think it looks good, but could you squash it into a single commit?

Thanks for all your work!

brycefisher:
Yes, no problem! Thanks for accepting this PR and helping me take my first steps in C++ :-)

SimonKagstrom:
8c4330d looks good! However, the merge request is for the WIP commit, so it needs to be updated.

brycefisher:
@SimonKagstrom not totally sure I understand what you mean. I updated the PR title to remove "WIP" -- is that what you were referring to?

Otherwise, since I forced pushed the squashed commit, there's no more WIPs in the commit messages. It looks like this PR would merge cleanly onto the HEAD of master at https://github.com/SimonKagstrom/kcov/commit/7fc8daaa51d4d4e8141f49a441cdb98d6bc0e0ad

Let me know what I can do here!

SimonKagstrom:
Sorry, I misunderstood github! All merged and good!

Thanks a lot for your contribution!

brycefisher:
Wahoo!! Thanks @SimonKagstrom

PR #173, opened 2017-01-10
Add `libiberty-dev` to ubuntu install help

brycefisher:
In order to use the `--verify` flag, the `libiberty-dev` package seems to be necessary when compiling kcov on some versions of ubuntu.

In my desktop version of ubuntu 16.10, its unnecessary. However, in the docker image `buildpack:yakkety`, the package is not installed by default.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -9,7 +9,7 @@ elfutils-devel and *not* elfutils-libelf-devel.
 
 Ubuntu
 ------
-Install binutils-dev libcurl4-openssl-dev zlib1g-dev libdw-dev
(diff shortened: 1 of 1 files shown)


SimonKagstrom:
Yes, makes sense. I didn't know they had split it out.

Is this the case in older Ubuntus though? I.e., does 16.04 and earlier have libiberty-dev or is it just part of binutils?

brycefisher:
Hmm, I've only checked on ubuntu 16.10. Not sure about ubuntu 16.04.

It's also possible that the docker images are fairly stripped down compared to regular ISOs you might download to run on a desktop or server.

SimonKagstrom:
I checked 16.04 and 14.04, and both have libiberty-dev, so the description is fine!

PR #176, opened 2017-01-13
ignore lines that are marked as unreachable

neumannt:
Sometimes source code is unreachable in reality, even though the compiler
is unable to infer that. With this patch the line can be marked as
unreachable, for example in a comment:

...
if (!someCall())
	exit(1); // unreachable, call never fails

When executing kcov as

kcov --unreachable=unreachable ...

That line is now ignored (unless it is reached), eliminating noise from the
coverage info.


SimonKagstrom:
Merging, but the Travis errors are strange. Travis builds are a bit unstable, but typically the entire run gets stuck (and most often on OSX).

Thanks for your contribution!

PR #181, opened 2017-01-26
Document simple parser

austin987:


SimonKagstrom:
Great! Thanks!

If it becomes popular, I guess we could make it a "proper" option, skipping the --configure= step.

Issue #192, opened 2017-04-10
Missing some bash stdout with self-built binary


astraw38:
Using kcov that was built on Ubuntu 14.04 (or CentOS 7.2.1511), I am missing some of the stdout data from the originally executed bash scripts:

```bash

[astraw][/home/astraw/workspace/gcc_test] $ /usr/local/bin/kcov . test.sh
begin
total 72
[astraw][/home/astraw/workspace/gcc_test] $ cat test.sh
#!/bin/sh

echo "begin"
/bin/ls -la
echo "end"
[astraw][/home/astraw/workspace/gcc_test] $ ls -la
total 72
drwxrwxr-x  4 astraw astraw 4096 Apr 10 10:16 .
drwxr-xr-x 14 astraw astraw 4096 Apr 10 10:00 ..
-rw-r--r--  1 astraw astraw  141 Apr 10 10:19 amber.png
-rw-r--r--  1 astraw astraw  101 Apr 10 10:19 bash-helper-debug-trap.sh
-rw-r--r--  1 astraw astraw   36 Apr 10 10:19 bash-helper.sh
-rw-r--r--  1 astraw astraw 2687 Apr 10 10:19 bcov.css
drwxr-xr-x  3 astraw astraw 4096 Apr  4 13:40 data
-rw-r--r--  1 astraw astraw  167 Apr 10 10:19 glass.png
-rw-rw-r--  1 astraw astraw 3818 Apr 10 10:19 index.html
-rw-rw-r--  1 astraw astraw  192 Apr 10 10:19 index.json
drwxr-xr-x  2 astraw astraw 4096 Apr  4 13:40 kcov-merged
-rwxrwxr-x  1 astraw astraw 7536 Apr  6 08:02 kcovtest
-rw-rw-r--  1 astraw astraw 1620 Apr  6 08:02 kcovtest.c
-rw-r--r--  1 astraw astraw 7504 Apr 10 10:19 libbash_execve_redirector.so
-rwxrwxr-x  1 astraw astraw  637 Apr  4 13:47 test2.sh
-rwxrwxr-x  1 astraw astraw   47 Apr 10 10:16 test.sh
[astraw][/home/astraw/workspace/gcc_test] $ 
```

This doesn't happen when using kcov installed from apt (but this *isn't* a solution for centos machines):

```bash
[astraw][/home/astraw/workspace/gcc_test] $ /usr/bin/kcov . test.sh
begin
total 72
[...]

PR #193, opened 2017-04-12
Handle all available bash output

astraw38:
Should fix #192

PR #211, opened 2017-07-20
bash-engine: Check stdout readable before getline

mbland:
Pull SimonKagstrom/kcov#193 (written to resolve SimonKagstrom/kcov#192) caused the [mbland/go-script-bash](https://github.com/mbland/go-script-bash) Travis build to fail. That build clones SimonKagstrom/kcov each time, and after that pull request, kcov began hanging after printing the number of Bats test cases on Linux, e.g.:

-  https://travis-ci.org/mbland/go-script-bash/jobs/246933167

This change resolves the issue by calling `file_readable` on standard output before calling `getline`.

Though I don't have a perfect understanding yet, it appears that Bash scripts that run for some amount of time and produce some amount of standard output eventually end up blocking when trying to write. Previously, when the parent `kcov` process would subsequently block in `getline`, the child process running the Bash script somehow wasn't able to resume, implying some sort of deadlock.

At any rate, the new `bash_handle_all_output` test case reliably reproduces the failure and validates the fix. It required adding a timeout capability to `KcovTestCase.do`.

I've also removed `@unittest.expectedFailure` from `bash_subshell`, as it isn't necessary and it confirms that the behavior implemented by SimonKagstrom/kcov#193 remains intact.

### DIFF
--- a/src/engines/bash-engine.cc
+++ b/src/engines/bash-engine.cc
@@ -352,37 +352,24 @@ private:
 	// Printout lines to stdout, except kcov markers
 	void handleStdout()
 	{
-		// Has input?
-		if (!file_readable(m_stdout, 0))
-			return;
-
 		char *curLine = NULL;
-		ssize_t len;
 		size_t linecap = 0;
 
-		len = getline(&curLine, &linecap, m_stdout);
-		while (len > 0)
+		while (file_readable(m_stdout, 0) && getline(&curLine, &linecap, m_stdout) > 0)
 		{
-			std::string cur(curLine);
+			/* Check for line markers to filter these away.
+			 *
+			 * For some reason, redirection sometimes give kkcov@..., so filter that
+			 * in addition to the obvious stuff
+			 */
+			size_t kcovMarker = m_bashSupportsXtraceFd ? -1 : std::string(curLine).find("kcov@");
 
-			if (!m_bashSupportsXtraceFd) 
+			if (kcovMarker != 0 && kcovMarker != 1)
 			{
-				/* Check for line markers to filter these away.
-				 *
-				 * For some reason, redirection sometimes give kkcov@..., so filter that
-				 * in addition to the obvious stuff
-				 */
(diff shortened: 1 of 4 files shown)


mbland:
Hmm, there's something decidedly funky about that failed Linux Clang build. Maybe just restart that one task?

Also, the reason the coverage went down slightly is because the new `handleStdout` implementation has fewer lines of code.

mbland:
And, duh, an `strace -f kcov ...` just now revealed the smoking gun: The xtraceFd (FD 782) buffer was getting full:

```
[pid 30363] <... read resumed> "550\n", 4096) = 4
[pid 30363] write(1, "550\n", 4550
)        = 4
[pid 30363] read(5,  <unfinished ...>
[pid 30366] <... write resumed> )       = 4
[pid 30366] write(782, "kcov@handle-all-output.sh@3@(( +"..., 38) = 38
[pid 30366] write(782, "kcov@handle-all-output.sh@3@(( i"..., 44) = 44
[pid 30366] write(782, "kcov@handle-all-output.sh@4@echo"..., 37
```

This is where it started to hang. So a buffer was getting filled, but it wasn't stdout. I'm actually going to update the last commit to include this message; guess it'll kick off some new builds as well.

mbland:
Wow, there's something seriously strange going on with that Linux Clang build. Wonder if it has to do with the Trusty update noted just before the logs:

> This job ran on our Trusty environment, which was updated on Wednesday, July 12th. Read all about it in the docs and take note that you can add group: deprecated-2017Q3 in your .travis.yml file to use the previous version.

SimonKagstrom:
The Clang issue is probably caused by me using clang-3.9 from a PPA. In the system-mode-ng branch, I've changed a few things around this, so the clang builds should work again once that's merged. Don't worry about that problem.

Anyway, thanks for the fix! Also compliments for the very detailed commit message and bug report!

mbland:
My pleasure! Thanks for merging, and for such a great tool! (I mean, c'mon, code coverage for Bash—who knew?) :smile:

Issue #212, opened 2017-07-31
SIGSEGV running Rust tests under kcov - regression


mthebridge:
I'm hitting an issue where running a Rust test binary under kcov segfaults, but running the binary natively is fine:

```
$ LD_LIBRARY_PATH=/opt/rust/.multirust/toolchains/stable-x86_64-unknown-linux-gnu/lib/ kcov --exclude-pattern=/opt/rust/.cargo --verify /data/mlt/rustproj/slog-extlog/slog-extlog-derive/target/cov/slog_extlog_derive-33e2f62c06364b1a/ /data/mlt/rustproj/slog-extlog/target/debug/deps/slog_extlog_derive-33e2f62c06364b1a kcov: 94 invalid breakpoints skipped in /opt/rust/.multirust/toolchains/stable-x86_64-unknown-linux-gnu/lib/libstd-f4594d3e53dcb114.so
kcov: Process exited with signal 11 (SIGSEGV) at 0x7ffff7105131
```

Obviously due to the way kcov works I can't run it in a debugger, but if I dump a core and load that in gdb I can see the segfault is somewhere in jemalloc in thread initialization.

Unfortunately I can't seem to generate a simple example for repro, and I can't share the underlying code, but running under kcov v33 works as expected, indicating this is a regression - and indeed  `git bisect` found that the commit at fault is e620da497dfcf93eefef6e582c68b68b9303f84c, which fixed #180.  Annoyingly, the reason I want to run off master and not from the v33 release is because I want to be able to exclude lines within my project!

I appreciate there's not a lot to go on here - let me know if there's any useful diagnsotics I can send!

SimonKagstrom:
Don't be so humble! You've even found the faulty commit! That said, it's a bit surprising that this particular commit causes a crash: it shouldn't change breakpoint setting very much.

Do you use the #180 feature? I.e., via LCOV_EXCL_START?

Is this consistent across every run, or does it sometimes work?

mthebridge:
Thanks for the speedy reply!  It doesn't matter whether I'm using `LCOV_EXCL_START` or not, and yes it's 100% consistent - always fails on master, always works with v33...

SimonKagstrom:
And the commit immediately before e620da4 (i.e.,860a32) always work? Sorry for doubting git bisect a bit ;-)

mthebridge:
Yes.  I just manually checked out those two revs and compared. 

On e620da4:
```
$ cargo kcov
[...]

PR #216, opened 2017-08-11
Minor "unused" fixes

Dridi:
Only tested on Fedora 27.

### DIFF
--- a/src/engines/bash-execve-redirector.c
+++ b/src/engines/bash-execve-redirector.c
@@ -55,7 +55,6 @@ int execve(const char *filename, char *const argv[], char *const envp[])
 			 strstr(startBytes, "/bin/dash") != NULL)) {
 		char **replacementArgv;
 		unsigned i = 0;
-		int rv;
 
 		// Find end
 		for (i = 0; argv[i]; i++)
--- a/src/engines/ptrace.cc
+++ b/src/engines/ptrace.cc
@@ -362,7 +362,7 @@ public:
 		// A signal?
 		if (WIFSTOPPED(status)) {
 			int sig = WSTOPSIG(status);
-			int sigill = SIGUNUSED;
+			int sigill = SIGSYS;
 
 			// Arm is using an undefined instruction, so we'll get a SIGILL here
 #if defined(__arm__) || defined(__aarch64__)


SimonKagstrom:
Merging, thanks!

The travis errors seems very unrelated to this. The clang build errors have been fixed in the system-mode-ng branch, so will be fixed when that is merged.

Dridi:
Speaking of the test suite, I'm having trouble "just" running it. I submitted a review request [1] to include `kcov` in Fedora, but for now the packaging doesn't run the test suite. We have offline build requirements and as such I can't use `travis/Makefile`. Should I open an issue regarding an offline-able test suite?

[1] https://bugzilla.redhat.com/show_bug.cgi?id=kcov

SimonKagstrom:
Nice to hear, thanks! I'm a fedora user myself (but a bit behind on versions).

As for the tests, it's not extremely complicated: The testsuite has an own cmake file (tests/CMakeLists.txt) which builds the binaries. The testsuite itself is then run with ```tests/tools/run-tests```, which also has some help:

```
$ tests/tools/run-tests
Usage: run-tests <path-to-kcov-binary> <kcov-output> <path-to-test-builds> <path-to-kcov-source>
```

So that should be pretty easy to use outside of travis. However, if I remember correctly, there are some tests which are x86-only, so running them on e.g., ARM would be problematic.


I also need to merge ```system-mode-ng``` before releasing the next version. The system-mode-ng branch is mainly lacking tests for now, and cleanup of the commits. There is also a regression, #212, which should be fixed before release.

Dridi:
Indeed, if I build kcov in `build/` and the test suite in `build-tests/` I get this:

```
$ ../tests/tools/run-tests ../build/src/kcov output . ..
x....Clang-only
u.....x....u....u............xx..xx.......................x.......u....
Ran 76 tests in 66.304s

OK (expected failures=7, unexpected successes=4)
```

As you can see [1] I already exclude most of the available architectures, so currently it's only x86 and ppc64 (big endian). Having the ability to skip tests on specific architectures should maybe be baked in the test suite, but I can do that for the whole test suite in the RPM spec too. I will have a closer look when I have time, because if it's shell I should be able to wrap my head around it.

I'm also curious about those unexpected successes :)

[1] https://dridi.fedorapeople.org/review/kcov.spec

SimonKagstrom:
Even better, the test suite is written in Python :-)

Anyway, I think its more of a build issue. The test suite already contains checks for OS (OSX or Linux), so I think adding architectures shouldn't bee too hard.

There are unfortunately some test cases which are a bit unstable due to timing. As for the unexpected successes, I'll take a look at that again.

SimonKagstrom:
In travis, there should now only be one unexpected successes (well, in system-mode-ng anyway), and that one is really unexpected. Or rather, the test case is slightly too limited for the bugs to really show.

Dridi:
> Even better, the test suite is written in Python :-)

Too bad, I'm rather of the shell persuasion... I think I will wait for kcov-34 to start fiddling with the test suite.

PR #229, opened 2018-01-09
build: Avoid packaging man dir with cpack

pyssling:
Packaging the man dir breaks installation on most distributions.
Avoid packaging normal man dirs. Conflicts may still occur on
non-standard installation paths (not /usr or /usr/local).


SimonKagstrom:
Thanks for the fix Nils!

As you can guess, I'm quite a bit of a newbie around cmake and cpack...

PR #239, opened 2018-03-16
ptrace: Don't rely on PTRACE_{GET,SET}REGS for Aarch64

### LINKED ISSUE #178: AArch64 support
Somewhere in between v31 and v33, support for AArch64 seems to have been dropped. Below is the compilation error:

```
[  6%] Building C object src/CMakeFiles/kcov_sowrapper.dir/solib-parser/lib.c.o
/builddir/kcov-32/src/solib-parser/lib.c: In function 'force_breakpoint':
/builddir/kcov-32/src/solib-parser/lib.c:95:3: error: #error Unsupported architecture
 # error Unsupported architecture
```

Even though this method was included in v31, it did compile. I suspect it wasn't used before.

Dridi:
With this patch kcov 34 builds fine on Fedora: https://koji.fedoraproject.org/koji/taskinfo?taskID=25734805

If any other downstream distributor is interested, this patch even applies cleanly on top of both the master branch and the v34 tag. Since I haven't got around running the test suite during the RPM build (yet), to my knowledge it compiles but may not work for this architecture. Someone should test this patch.

Summoning @NoctuaNivalis

Fixes #178

### DIFF
--- a/src/engines/ptrace.cc
+++ b/src/engines/ptrace.cc
@@ -19,6 +19,10 @@
 #include <sys/types.h>
 #include <dirent.h>
 
+#if defined(__aarch64__)
+#  include <sys/uio.h>
+#endif
+
 #include <map>
 #include <unordered_map>
 #include <list>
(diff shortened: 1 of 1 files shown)


SimonKagstrom:
Thanks a lot!

I should say that Aarch64 is actually untested, so might not work.

PR #243, opened 2018-04-27
ptrace: include elf.h for NT_PRSTATUS

Ekleog:
In order to build on NixOS I had to add the following include on top of https://github.com/SimonKagstrom/kcov/pull/239.

`<elf.h>` was the least terrifying header in which `NT_PRSTATUS` appears defined in glibc 2.26-131. `<linux/ptrace.h>` had a more appropriate name, but being in the `<linux/` namespace I thought `<elf.h>` was a better pick.

Logs [before](https://logs.nix.ci/?key=nixos/nixpkgs.39624&attempt_id=0342d4c4-c880-47a9-b125-1bed464b92e1) and [after](https://logs.nix.ci/?key=nixos/nixpkgs.39624&attempt_id=49ea1849-53ca-4e37-bfed-177fb427ce7d), I don't have much more than that as testing goes :)

### DIFF
--- a/src/engines/ptrace.cc
+++ b/src/engines/ptrace.cc
@@ -21,6 +21,7 @@
 
 #if defined(__aarch64__)
 #  include <sys/uio.h>
+#  include <elf.h>
 #endif
 
 #include <map>


SimonKagstrom:
Merged, thanks! (Been travelling without computer a few days, hence the delay!)

Ekleog:
Thanks! (there are *much* longer-to-react maintainers than you are, no problem with 2 days ^^)

SimonKagstrom:
By the way, did you actually try it on Aarch64? It was implemented "blindly", so it's unclear if it actually works.

Dridi:
I hope to get a stab at the test suite at some point. Running it at packaging time should un-blind us ;-)

SimonKagstrom:
Nice!

The test suite has been somewhat x86-only, but I think those things should be fairly easy to fix.

Dridi:
Ideally I should be able to build with `make` and right after run the test suite with `make check`. I don't remember the details but it was a bit more involved so I stuck to "works on my machine" manual testing :(

SimonKagstrom:
Well, the kcov test suite uses a separate cmake step to build the test binaries. Perhaps the easiest thing is to check in `travis/Makefile`, which builds and runs the testsuite.

The same thing goes for the unit tests (which are unfortunately a bit too few right now).

Ekleog:
The issue is that `travis/Makefile` isn't usable as-is, as it requires `sudo` to work correctly.

I had looked quickly into running the test suite as part of the build phase, but haven't managed to have it pass even on x86_64 for the time being, with:
```bash
# (starting in <kcov sources>/build)
patchShebangs ../tests/ # fixes up things like #!/bin/sh usage into reproducible nix-store paths
mkdir build-tests

cd build-tests
cmake ../../tests
cd ..

make build-tests
../tests/tools/run-tests ./src/kcov /tmp/ ../build-tests/ ./ -v
```

[(full package with which I'm testing)](https://gist.github.com/Ekleog/bfe1e358eb1f982d878b09195d3a910b)

I'm getting a few assertion errors and file-not-found errors and errors like:
```
runTest (compiled.collect_no_source) ... cp: cannot stat './/tests/short-file.c': No such file or directory
gcc: error: ../build-tests//main.cc: No such file or directory
[...]
```

Not having much time to debug that for now, just putting it here in case it can help :)

SimonKagstrom:
Yes, the travis Makefile should only be used to look in, it's quite specific to the travis environment :-)

Is it built in the source directory? I don't think that should be a problem as such, but in general it's not recommended.


Anyway, for distros, I think it makes sense to run a smaller test suite to verify basic functionality. Perhaps just building one of the tests manually (i.e., `gcc -g ...`) and verifying that `output/cobertura.xml` shows covered lines.

Looking at the test suite, there are still x86-specific parts (illegal insn, the 32-bit stuff etc), which will not work on non-x86-platforms.

Ideally, I'd like a travis-for-ARM, a travis-for-PPC and a travis-for-Aarch64 to test these during CI, but I guess that isn't going to happen...

Dridi:
> Ideally, I'd like a travis-for-ARM, a travis-for-PPC and a travis-for-Aarch64 to test these during CI, but I guess that isn't going to happen...

I'd do that the other way around: having the test suite aware of arch-independent vs archful tests and enabling the latter only on a compatible system.

SimonKagstrom:
I agree, it should be done that way. OSX/Linux is handled similarly. But what I really meant was a Travis which runs on ARM etc CPUs, so that all supported architectures are tested with each push.

Dridi:
Oh but once you have it set up we can figure how to set this up. At the very least it should be possible to manually build on _all_ architectures supported by Fedora with a single push (that's how I test archful packages before submitting them).

SimonKagstrom:
Thanks!

Well, for starters, it *should* be possible to build and the tests without the travis stuff. This is what I do:

```
mkdir -p build/kcov build/tests
mkdir src
cd src && git clone <kcov>
# build kcov with cmake + make itself in build/kcov
cd build/tests && cmake ../../src/tests && make # build test programs

cd src
tests/tools/run-tests ../build/kcov/src/kcov /tmp/kcov ../build/tests/ . -v
```

Basically at least.

PR #244, opened 2018-04-30
Various CMake and CPack fixes/improvements.

jamiesnape:
PTAL @SimonKagstrom. We could simplify the build further if we required CMake 3.0 (circa 2014) rather than 2.6 (circa 2008).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,11 @@ cmake_minimum_required (VERSION 2.6)
 # ====================================
 project (kcov)
 
+if(POLICY CMP0042)
+  # MACOSX_RPATH is enabled by default.
+  cmake_policy (SET CMP0042 NEW)
+endif()
+
 if(NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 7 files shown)


SimonKagstrom:
Wow, thanks!

I guess it shows that I'm a bit of a CMake newbie still :-).

Anyway, I've had users who build kcov on pretty old systems (e.g., without C++11), so I think 3.0 might still be too fresh for a few of my users.

jamiesnape:
Thanks for merging that. A quick follow up, looks like you are using language features that were not introduced until 2.8.0 (`$<TARGET_FILE:>`) and properties that were not introduced until 2.8.9 (`POSITION_INDEPENDENT_CODE`). Not difficult to fix, but CMake 2.6 users would run into errors.

SimonKagstrom:
OK, thanks. So I guess the requirement should really read >= 2.8.9? I haven't had complaints, so I think the de facto requirement should be OK.

SimonKagstrom:
OK, that would be great, thanks!

Issue #248, opened 2018-05-13
The bash engine can hang if the traced program outputs data to stdout without newlines


SimonKagstrom:
See the discussion in #191

SimonKagstrom:
Fixed, I believe, with ad17136.

SimonKagstrom:
I don't have your project at hand right now, but if your bash supports xtrace FD (I think version 4.2 and above), I would have thought the output would be the same as without kcov. Is it not so?

ishbguy:
Normal output:
```bash
[ishbguy@arch 14:46 2 0 build]master$ bash ../../baux/test/self/01-test-baux.sh
PASS  1 subtest: test die
PASS  2 subtest: test die_hook
PASS  3 subtest: test warn
PASS  4 subtest: test version
PASS  5 subtest: test proname
PASS  6 subtest: test usage
PASS  7 subtest: test import
TOTAL: 7, PASS: 7, FAIL: 0, SKIP: 0, 100% pass.
```
Run with `kcov`:
```bash
[ishbguy@arch 14:46 2 0 build]master$ ./src/kcov /tmp/kcov ../../baux/test/self/01-test-baux.sh
        die_hook() { echo '\''I am in die_hook'\''; }
        run_ok '\''$status -eq 1 && $output =~ die_hook'\'' die
    }'
I am in die_hook =~ die_hook ]]
        run_ok '\''$status -eq 1 && $output == test'\'' warn test
        warn_two() { warn test; warn test; }
        run_ok '\''$status -eq 2'\'' warn_two
    }'
        local VERSION=
        run_ok '\''$status -eq 1 && $output =~ VERSION\ variable'\'' version

        local VERSION='\''v0.0.1'\''
        run_ok '\''$status -eq 0 && $output =~ v0.0.1'\'' version
    }'
        local PRONAME=proname
        run_ok '\''$output == proname'\'' proname
        PRONAME=
        run_ok '\''$status -eq 0 && $output =~ 01-test-baux.sh'\'' proname
    }'
        local HELP=
        local VERSION=
        run_ok '\''$status -eq 2 && $output =~ HELP && $output =~ VERSION'\'' usage
        local HELP='\''usage help'\''
        local VERSION='\''v0.0.1'\''
        run_ok '\''$status -eq 0 && $output =~ v0.0.1 && $output =~ usage\ help'\'' usage
    }'
You need to define a HELP variable. =~ HELP ]]
[...]

Issue #191, opened 2017-04-05
travis: The setpgid_kill test is unstable


SimonKagstrom:
Sometimes hangs, probably waiting for a result which has already arrived or so.

ishbguy:
I encounter the same situation then running kcov after  travis build, kcov hang without output for a long time, and finally killed by travis. For details you can see: https://travis-ci.org/ishbguy/baux/builds/376708953

SimonKagstrom:
OK, interesting - and this is with bash as well. The `setpgid_kill` test has been stable for a while though, so it appears like I've managed to avoid a race somehow.

Does this happen if you run it on your computer as well?

ishbguy:
yes, I encounter the same situation when running `kcov` on my local machine , so I need to kill kcov process to get back my shell controling. I try to use `--bash-method=DEBUG` option, `kcov` can run through, but the coverage report shows a **low percentage (lower than 10%)** on my code, which I think `kcov` collects not enough datas.

SimonKagstrom:
Yes, sounds like that. Can you run with --debug=31 and see if you can see
where it hangs? Alternatively, run kcov via gdb (gdb --args kcov [other
options]) and break it when it hangs)

ishbguy:
I have tried to run `kcov --debug=31`, and this is the out put:
[kcov-debug.txt](https://github.com/SimonKagstrom/kcov/files/1992192/kcov-debug.txt)

ishbguy:
I use `lldb` to debug, and I was hanged in the `handleStdout`'s `getline` calling, a part of the debug output is in the txt below:
[lldb-kcov-debug.txt](https://github.com/SimonKagstrom/kcov/files/1992515/lldb-kcov-debug.txt)

SimonKagstrom:
Hmm.. That's the one which reads from the covered programs output. Do you need to use `--bash-method=DEBUG`, and does it make a difference without it?

Does running bash with `-x` work as expected?

ishbguy:
1. running `kcov --bash-method=DEUBG dir test-script.sh` is OK, but `REPORT hit` little code.
2. running `kcov dir test-script.sh` will hang.
3. running `bash -x test-script.sh` is OK.

SimonKagstrom:
OK! Thanks for the analysis, this should be not too difficult to fix! Perhaps a pull-request even?

[...]

PR #249, opened 2018-05-14
Fix bash engine hanging

ishbguy:
The bash engine can hang if the traced program outputs data to
stdout without newlines.

Reference issue #248 and #191 (discussion)

### DIFF
--- a/src/engines/bash-engine.cc
+++ b/src/engines/bash-engine.cc
@@ -121,7 +121,7 @@ public:
 				return false;
 			}
 			m_stdout = fdopen(stdoutPipe[0], "r");
-			if (!m_stderr) {
+			if (!m_stdout) {
 				error("Can't reopen the stdout pipe");
 				fclose(m_stderr);
 
@@ -145,7 +145,7 @@ public:
 			    return false;
 			}
 			if (dup2(stdoutPipe[1], 1) < 0) {
-			    perror("Failed to exchange stderr for pipe");
+			    perror("Failed to exchange stdout for pipe");
 			    return false;
 			}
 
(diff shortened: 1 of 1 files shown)


SimonKagstrom:
Thanks!

However, the testsuite now fails, which I needs to be investigated. As for the implementation, doesn't this fail for lines longer than 10 characters? Or would that perhaps be handled in multiple prints?

The failing test can be run as

```kcov --debug-force-bash-stderr /tmp/kcov tests/bash/redirect-stderr.sh```

and should only output `stuff` and nothing else. After your fix, I think some of the kcov lines gets mixed in there.

**Comment by ishbguy** [comment]
I fixed typo at the same time, I will recover these 2 lines and try again.
```git-commit
| @@ -121,7 +121,7 @@ public:
|                               return false;
|                       }
|                       m_stdout = fdopen(stdoutPipe[0], "r");
| -                     if (!m_stderr) {
| +                     if (!m_stdout) {
|                               error("Can't reopen the stdout pipe");
|                               fclose(m_stderr);
|
| @@ -145,7 +145,7 @@ public:
|                           return false;
|                       }
|                       if (dup2(stdoutPipe[1], 1) < 0) {
| -                         perror("Failed to exchange stderr for pipe");
| +                         perror("Failed to exchange stdout for pipe");
|                           return false;
|                       }
|
```

ishbguy:
Hi @SimonKagstrom , 
I found that the original logic is to get a whole line at once, my solution is to get a sub-string at once, which will make the remain logic to ignore  the first sub-string containing `kcov@` or `kkcov@`,  except `kkkcov@`, `kkkkcov@` , etc, and the remain sub-strings of this line will also be printed out. 
Though, I use the command `kcov --debug-force-bash-stderr /tmp/kcov tests/bash/redirect-stderr.sh` to test on local machine, the output does not contain `kcov@` , so I have no idea why the travis test failed.

ishbguy:
As for above description, my solution devides a line into many small sub-string, I try to contruct these small parts to become a line again, then I found that I lost the first sub-string in most case:
```C
void handleStdout()
{
    char line[1024];
    char *line_ptr = &line[0];
    size_t len = 10 +1;  // string with null

    while (file_readable(m_stdout, 0) && fgets(line_ptr, len, m_stdout) == line_ptr) {
        // read remain string at once if the first sub-string is not a line
        if (strchr(line_ptr, '\n') == NULL) {
            line_ptr = &line[len-1]; // next time will voerride the null char
            len = 1024 - len; 
            continue;
        }
        size_t kcovMarker = m_bashSupportsXtraceFd ? -1 : std::string(line).find("kcov@");
        if (kcovMarker != 0 && kcovMarker != 1) {
            printf("%s", line);
        }
        // restart from 0
        line_ptr = &line[0];
    }
}
```  
I guess after the first `fgets`, `m_stdout` became unreadable, so I lost the first sub-string.
Now, I think the `kcov` sometimes hanging is possiblly because the `m_stdout` become unreadable while `getline` waiting for a `\n`. 
For my solution can not completly solve this problem, I will close this PR :( .

SimonKagstrom:
How can I run your test case / project on my own machine? I might be able to take a look at it myself.

ishbguy:
All my test cases are in the my `baux` project's `test/self` directory with `.sh` suffix, you  can easily test it as a nomarl shell script, like: `kcov /tmp/kcov test/self/00-test-baux-test.sh`.

SimonKagstrom:
OK, good. I'll be travelling this weekend, so won't look at it for some time.

SimonKagstrom:
I've looked a bit more on it now, and this comes from the use of PS4 in kcov.

You can get the same output effect by running your program with `bash -x <args>`, which is basically what kcov uses. If I run kcov with the DEBUG trap instead, I don't see these, i.e.,

```
kcov --bash-method=DEBUG <same-args-as-before>
```

but the debug trap has other limitations, which may perhaps affect you.

Issue #135, opened 2016-03-03
Covering binaries on OS X


zbeekman:
Will this build and install on OS X? I'm guessing the answer is "no" because ELF/DWARF is Linux specific? (Please forgive my crude understanding of everything deeper than compiled language source code... I am but a lowly aerospace engineer/computational scientist with no formal CS background)

SimonKagstrom:
Technically, the answer is "yes" :-). However, on OS X, only Python and Bash coverage collection works. And as you guess, the ELF/ptrace parts are Linux-specific. DWARF is used on OS X as well, so a port is certainly possible, but will require a bit of work:

- kcov uses libdw from elfutils for the DWARF parsing, which is closely tied to ELF
- the ptrace parts will be similar, but the interface is not exactly the same, so this will have to be split
- a macho-parser.cc (how's that for a name!) that implements the needed parts from elf-parser.cc will have to be created

Unfortunately, I don't run OS X myself (although I use Apple hardware, but with Linux), so I don't really have a time frame for a full port to OS X.

zbeekman:
Great thanks for clarifying!

One more follow up question: If I only want to generate shell script coverage data, are there features in kcov that are missing from shcov? It seems that building kcov is a little bit more involved, and I was wondering if the feature set for shell script coverage info is the same between the two.

Not having to compile with gcov flags is an awesome feature of kcov, so I was considering dropping the "Coverage" build from our Travis-CI test matrix and using kcov to get coverage for the release builds. It sounds like I can do this on the Linux VMs but the OS X VMs will still need to  use gcov for coverage data.

zbeekman:
P.S. thanks for the cool project!

SimonKagstrom:
Thanks!

Yes, shcov is quite a bit more limited than kcov: It doesn't support coveralls, has more basic HTML output, can't accumulate data between runs, no filtering and a bit more false positives/negatives for bash.

I'll keep this bug report open for the OS X collection, but don't expect to get much work done on it for a while!

SimonKagstrom:
By the way: If you're using Travis-CI for Bash collection on OS X, the travis/Makefile file in the kcov distribution contains the needed homebrew packages for the kcov build:

```
brew update
brew install homebrew/dupes/zlib
brew install python
brew install bash
```

zbeekman:
Thanks. So I take it that kcov requires bash 4?

SimonKagstrom:
[...]

PR #253, opened 2018-05-15
FreeBSD support

asomers:
Add FreeBSD support to kcov.  It has all the capabilities of Linux except for full-system instrumentation (I haven't tried to port that, because I don't personally need it).  The ptrace module eschews CPU pinning.  It can be implemented, but I'm not sure it's necessary.  I didn't implement liblldb support either, though it should be possible.  All tests pass.


asomers:
I can't explain why the coverage of ptrace_linux.cc is so low.  That file is definitely being exercised; otherwise all of the compiled tests would fail.

asomers:
Nevermind, I do understand it.  The run-tests-linux-gcc target in travis/Makefile doesn't run kcov on itself while doing anything useful.  It just runs kcov on itself while producing the help text.

SimonKagstrom:
Wow, awesome stuff! Thanks a lot for doing this!

Is there something similar to Travis-CI with FreeBSD? Would be nice to know that it hasn't been broken by other changes.


At least on Linux, ptrace is not recursive, so kcov can't instrument itself. So the travis tests only do self-coverage for Python/Bash.

The liblldb stuff is for OSX. I haven't found any good libraries for parsing Mach-O binaries, so it simply uses the debugger as a library to implement both parsing and an "engine". It's quite a bit slower than Linux and presumably FreeBSD though, so the native approach is much better.

The system-mode is probably not relevant for non-Linux systems, and also isn't quite stable anyway (pending time for debugging).

asomers:
Sadly, there's no free hosted CI system for FreeBSD.  Pretty much the only options are to setup your own BuildBot server or to setup your own Jenkins server.  Have you looked at Goland?  It's a debugger for Go.  It uses ptrace on OSX, so it must've found a way to dump OSX binaries.

SimonKagstrom:
OK, too bad. I have the same problem for PowerPC/ARM, which might well be broken as I don't regularly run on such systems.

I haven't looked at Goland, but I did take a look at Chromium Breakpad which parses Mach-O binaries (see the discussion in #135). However, the LLDB solution is simple and works well (apart from slowness), so I decided I wasn't up for the "proper" job.

kimkim19642004:
Hello everyone, **kcov** is really useful !!

I am using **kcov** to test my bash scripts under the FreeBSD environment shown below:

| Difficulty (highest first) | Environment | kcov working status |
| --- | --- | --- |
| 1 | [Cirrus CI](https://cirrus-ci.org/) FreeBSD (version 11/12) | The message 'kcov: error: Can't find or open <file>' was displayed. See [Error logs](https://cirrus-ci.com/github/kimkim19642004/po2bats-autotools). |
| 2 | GhostBSD (version 19) on my old Laptop-PC | 'kcov' has no problem. Desktop-system is unstable. |
| 3 | FeeBSD (version 12) on my old Laptop-PC | The message 'kcov: error: Can't find or open <file>' was displayed once, but I have not been able to reproduce the phenomenon. |
| 4 | GhostBSD (version 18) on my old Laptop-PC | No problems |

#### TODO:
- I want to get the green badge!! of [Cirrus CI](https://cirrus-ci.org/).
- Test and code review for [configuration.cc](https://github.com/SimonKagstrom/kcov/blob/master/src/configuration.cc).
- Clearly separate the processing for **'optional arguments'** and **'file operands'**. See [the sample codes of getopt(3)](https://man.cx/getopt(3)).

#### Reference
- [Error logs](https://cirrus-ci.com/github/kimkim19642004/po2bats-autotools)
- [.cirrus.yml](https://github.com/kimkim19642004/po2bats-autotools/blob/master/.cirrus.yml)
- [debug main.cc](https://github.com/kimkim19642004/po2bats-autotools/blob/master/main.cc)
- [debug configuration.cc](https://github.com/kimkim19642004/po2bats-autotools/blob/master/configuration.cc)

asomers:
Try invoking pobats.sh by absolute path.  Also, this issue isn't really the right place to discuss your bug.  You should open a new issue if you think there's a bug in kcov.

PR #258, opened 2018-05-30
Fix false bash line count on function with spaces between name and parenthesis

eugpermar:
```fn()``` was not evaluated as executable line, but ```fn ()``` was.

### DIFF
--- a/src/engines/bash-engine.cc
+++ b/src/engines/bash-engine.cc
@@ -647,6 +647,11 @@ private:
 			// fn() { .... Yes, regexes would have been better
 			size_t fnPos = s.find("()");
 			if (fnPos != std::string::npos) {
+				// Trim valid last spaces between function name and parenthesis
+				while (fnPos > 1 && ' ' == s[fnPos - 1]) {
+					--fnPos;
+				}
+
 				std::string substr = s.substr(0, fnPos);
 
 				// Should be a single word (the function name)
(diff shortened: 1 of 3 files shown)


eugpermar:
Hi! Could you please guide me in order to create proper unit testing in kcov for bash file?

Thanks and regards!

SimonKagstrom:
OK, thanks for the fix!

The travis build errors is from a commit I forgot to revert.

The bash stuff doesn't have proper unit tests, although I guess it could be useful. It's tested via system tests, which are the stuff in `tests/tools/bash.py`. These tests can be run with `tests/tools/run-tests`, which also explains how it should be run.

For actual unit tests, you need `crpcut` and `trompeloeil` (the latter is added as a submodule in kcov), but the bash engine isn't really written to be unit test friendly unfortunately.


I think your particular error is easy to add to the system tests though.

eugpermar:
Hi! I finally managed to run only bash tests so I hadn't had to deal with other kinds of test. Waiting for your review!

SimonKagstrom:
Thanks!

The patch looks good to me, but the new test fails during testing:

 https://travis-ci.org/SimonKagstrom/kcov/jobs/389345999#L4484

Does it work when you run it manually?

eugpermar:
I forgot to add the test file to the repo, and I assumed the travis failure was because of the previous ones. Amended & pushed!

SimonKagstrom:
Merged, thanks a lot!

eugpermar:
You're welcome!

Issue #210, opened 2017-07-20
Generated files with `.json` extension do not conform to the JSON file format


chradcliffe:
The `.json` files generated by `kcov` contain variable definitions and therefore are not well-formed JSON files. These files are included in `<script>` tags as content-type "text/javascript" and so are evidently not meant to be pure JSON.

When used with a web server that sets `X-Content-Type-Options: nosniff` and sends files with the `.json` extension with content-type `application/json`, modern browsers will refuse to execute the contents of the `.json` file and the `kcov` report will not render properly.

I can see two possible resolutions to this issue:

1. Rename generated `.json` files to `.js`. This is probably the least-effort solution.
1. Modify the `.json` files to be proper JSON objects and reference members of a json object instead of the individual variables set inside the `.json` files.

I can submit  a PR once the right solution is decided upon.

chradcliffe:
Here is a Python web server script that sets the `X-Content-Type-Options: nosniff` header:

```python
#!/usr/bin/env python3

from http.server import HTTPServer, SimpleHTTPRequestHandler

class Handler(SimpleHTTPRequestHandler):
    def do_GET(self):
        self.send_header("X-Content-Type-Options", "nosniff")
        super().do_GET()

server_address = ('', 8000)
httpd = HTTPServer(server_address, Handler)
httpd.serve_forever()
```
If you run this script with the current working directory as the root of the kcov output tree, the served pages will fail to render properly under at least: Chrome 59.0.3071.115 (Windows), IE7 (Windows), and MS Edge (Windows).

SimonKagstrom:
I've been away on computer-less vacation for a while, so sorry for not responding.

Anyway, thanks for the report - I guess it shows I'm not really a JavaScript and JSON newbie, but a JS/JSON infant! I don't actually remember why the variables were introduced in the output, but at least the "Not really json, but anyway" comment shows that I at least was aware of it being a problem.

Solution 2 sounds more clean (being less web-browser centric), or is there some additional advantage of having it in JS format?

Dridi:
I don't know about advantages of having a js format, but I was hoping to manipulate the json (with `jq` or something else) to ignore lines that should always have 0 hits (for example looking for `WRONG` macros in the Varnish code base, or `unreachable!` macros in Rust).

[...]

PR #260, opened 2018-06-11
Code coverage report behind nginx/1.15.0 can not show well which respose index.json as application/json

lisongmin:
The develop console output：

```
Loading failed for the <script> with source “https://.../master/test_sh/index.json”
ReferenceError: data is not defined[Learn More] kcov.js:41:3
```

When I check the response data， an error output:

```
SyntaxError: JSON.parse: unexpected end of data at line 1 column 1 of the JSON data
```

Change file name to .js fix this problem. Cloud we change the file extension to fit the real file type?

### DIFF
--- a/data/index.html
+++ b/data/index.html
@@ -10,7 +10,7 @@
 <font color=red><B>ERROR:</B></font> JavaScript need to be enabled for the coverage report to work.
 </noscript>
 
-<script type="text/javascript" src="index.json"></script>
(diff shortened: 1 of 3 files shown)


lisongmin:
The change may effect to json writer which is not expected. 

I will try to avoid this by more code ...

SimonKagstrom:
Yep, that's a good change! Issue #210 would also be nice to have solved, and I guess we could rename them back again when that is done.

PR #268, opened 2018-09-21
debian package files

pauldreik:
Hi, thanks for the nice talk at SwedenCpp yesterday!
I made some rudimentary packaging files for debian so it is easy
to (un-)install kcov. Tested on Ubuntu 18.04 and Debian testing ("Buster").
I know there is an upstream package, but that seems outdated.


SimonKagstrom:
Wow, thanks a lot on both accounts!

Yes, the other Debian package is quite outdated unfortunately.

PR #282, opened 2019-02-01
Adding uninstall CMake target

akossomfai:
Adding uninstall CMake target. Installation now can be removed using "make uninstall".


SimonKagstrom:
Great, thanks a lot!

PR #286, opened 2019-03-06
Do not require XCode to be installed on macOS

jamiesnape:
Manually fixing up the RPATH should not really be necessary, but I have not had a chance to investigate that. The previous fix caused an unnecessary dependency on (a hard-coded path to) XCode.

PTAL @SimonKagstrom.


SimonKagstrom:
Merged, thanks a lot!

PR #296, opened 2019-05-21
Full coverage-04.dtd compliance

BotellaA:
There was missing attributes to fully satisfy the format.


SimonKagstrom:
Thanks a lot for digging into this!

PR #303, opened 2019-07-07
Enable functrace

### LINKED ISSUE #234: kcov can't cover code in function when using --bash-method=DEBUG with recent versions of bash
I'm working on a Ubuntu14.04 OS, kcov gave me something weird.<br>

Saying that there is a bash script named `a.sh`:<br>
```
#!/bin/bash
testfunc(){
    echo 'number:'
}

testfunc
```

Then I executed the command as below:<br>
addparm-after-upgrade

However, it got the result:<br>

which should be 100%.

However, if I modify the `a.sh` as below, kcov will give 100% to me:<br>
```
#!/bin/bash
echo 'number:'
```

Why doesn't kcov work as expected if there are some functions in the script?

ko1nksm:
Enable `functrace` shell option to cover in function when using `--bash-method=DEBUG`

Require this option to inherit DEBUG trap in shell function. `functrace` was introduced in bash 3.0. 

I think this PR can close #234.

And I found probably unnecessary leading space.

### DIFF
--- a/src/engines/bash-helper-debug-trap.sh
+++ b/src/engines/bash-helper-debug-trap.sh
@@ -1,4 +1,4 @@
- #!/bin/sh
+#!/bin/sh
 
 set -o functrace
 trap 'echo "kcov@${BASH_SOURCE}@${LINENO}@" >&$KCOV_BASH_XTRACEFD' DEBUG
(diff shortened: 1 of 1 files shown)


SimonKagstrom:
Ah, great! Thanks a lot for that! #234 has been an irritating bug.

PR #304, opened 2019-07-07
Parse correctly the tracing line of PS4

ko1nksm:
This PR fixes unexpected output to stderr when using `--bash-method=PS4`

[test.sh]
```sh
#!/bin/sh

a='line1\'\''
line2'

if [ "$a" ]; then
  :
fi
```

```console
$ kcov /tmp/cover ./test.sh >/dev/null
line2'
line2' ']'
```

I think it works correctly but there is no test. I do not know where to write the test. How can I check if it is broken?

By the way, I found this problem during develop [shellspec](https://github.com/shellspec/shellspec) - new BDD testing framewrok for shell script. Unexpected output to stderr caused failure the tests via shellspec. I managed to integrate kcov using `--bash-method=DEBUG` (with workaround for #303). kcov is a great software! thanks!

### DIFF
--- a/src/engines/bash-engine.cc
+++ b/src/engines/bash-engine.cc
@@ -256,19 +256,18 @@ public:
 		std::string cur(curLine);
 		// Line markers always start with kcov@
 
-		size_t kcovStr = cur.find("kcov@");
-		enum InputType ip = getInputType(cur);
-		if (kcovStr == std::string::npos)
-		{
-			if (m_inputType == INPUT_NORMAL)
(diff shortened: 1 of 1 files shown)


SimonKagstrom:
OK, thanks a lot for the fix!

As for writing tests, the tests/tools directory contains the test suites. This should proabably go to the bash.py test file, and the example script itself added to a new file in tests/bash. To run the test suite, run tests/tools/run-tests and follow the on-screen usage (basically you need to point to where kcov was compiled and the kcov source directory).

Issue #305, opened 2019-08-07
Python 2 EOL


Dridi:
Hi,

With Python 2 going EOL soon, it will become impossible to build the test suite starting with Fedora 31 without patching the source tree. Fedora switched /usr/bin/python to python3, so existing scripts (like `tests/multi-fork/generate-functions.py`) are starting to fail.

I think shebangs should move from the un-versioned `#!/usr/bin/env python` to an explicit `#!/usr/bin/env python2` and ideally ported to Python 3 immediately for simple scripts to eventually drop Python 2 support altogether.

Depending on your interest, I may open a pull request to version the shebangs, and maybe move trivial scripts to Python 3.

SimonKagstrom:
Ah! Yes, the test scripts are probably pretty clumsily written (I'm very much not a Python expert!), so I guess that should be done.

In the short term, I think you python2-shebang is probably best, and then the scripts can be converted one by one.

Dridi:
Shebangs are done in #306. With this I can run the test suite on the upcoming Fedora 31.

On Fedora 30 I get the following redacted results:

```
[...]
runTest (compiled.address_sanitizer_coverage) ... Clang-only
unexpected success
[...]
runTest (bash.bash_coverage_debug_trap) ... unexpected success
[...]
runTest (system_mode.system_mode_can_record_and_report_binary) ... x86_64
ERROR
unTest (system_mode.system_mode_can_start_and_stop_daemon) ... ok
runTest (compiled.vfork) ... ok

======================================================================
ERROR: runTest (system_mode.system_mode_can_record_and_report_binary)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/dridi/src/kcov/tests/tools/system_mode.py", line 65, in runTest
    dom = parse_cobertura.parseFile(testbase.outbase + "/kcov-report/main-tests/cobertura.xml")
  File "/home/dridi/src/kcov/tests/tools/parse_cobertura.py", line 46, in parseFile
    dom = xml.dom.minidom.parseString(readFile(filename))
  File "/home/dridi/src/kcov/tests/tools/parse_cobertura.py", line 9, in readFile
    f = open(name,'r')
IOError: [Errno 2] No such file or directory: '/tmp//kcov-report/main-tests/cobertura.xml'

----------------------------------------------------------------------
Ran 85 tests in 80.653s

FAILED (errors=1, expected failures=7, unexpected successes=2)
```

[...]

PR #306, opened 2019-08-09
Avoid unversioned python shebangs

Dridi:
Starting with Fedora 31 plain python means Python 3 and it wouldn't be
surprising that other operating systems gradually make the same change
as Python 2 is going EOL.

Refs #305


SimonKagstrom:
Great, thanks a lot!

PR #307, opened 2019-08-11
Convert tests scripts to Python 3

Dridi:
It brought some warnings that I also fixed.

Refs #305


Dridi:
It failed on Travis CI:

```
runTest (bash.bash_coverage_debug_trap) ... unexpected success
```

It also unexpectedly succeeds on my Fedora 30 system.

SimonKagstrom:
Well, I guess that test should be expected to succeed for now. I don't really remember why it wasn't working before, but as long as both Travis and Fedora passes it I think it should be OK.

PR #308, opened 2019-08-13
 Convert src scripts to Python 3 

### LINKED ISSUE #305: Python 2 EOL

Dridi:
Both scripts were already python3-ready so only their shebang changes
here. In particular, Python 2 support in python-helper.py was not yet
removed since Python 2 is not EOL quite yet.

Fixes #305

### DIFF
--- a/src/bin-to-c-source.py
+++ b/src/bin-to-c-source.py
@@ -1,6 +1,4 @@
-#!/usr/bin/env python2
-
-from __future__ import print_function
+#!/usr/bin/env python3
 
 import sys, struct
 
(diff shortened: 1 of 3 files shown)


SimonKagstrom:
Super! Thanks a lot for your work!

PR #309, opened 2019-08-13
Fix the encoding issue for Non English in reports.

zhenweiding:
Add the utf-8 charset meta info for html reports.


SimonKagstrom:
Great, thanks!

sorry for not having noticed this :-)

PR #310, opened 2019-08-19
Update for latest debian stable (buster) docker image

ko1nksm:
Feel free to use it if you want.

- Use debian:stable-slim for builder.
- Remove --fix-missing pkg-config (unnecessary?).
- Update to python3 and libcurl4.
- Do not copy libs (reduce image size 203MB -> 117MB).
- Add /src/build to .dockerignore to do not add unnecessary files.
- Fix typo (/usr/local/chare/doc/kcov -> /usr/local/share/doc/kcov).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 Dockerfile
+build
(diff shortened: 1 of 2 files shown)


SimonKagstrom:
Good stuff, especially for a Docker newbie like me! Thanks a lot!

Issue #321, opened 2020-05-01
make kcov available in conan package manager


davidtazy:
it would be nice to add this tool as a "build-requires" in a conanfile.txt

SimonKagstrom:
I agree. I'm not personally knowledgeable in conan though, but perhaps someone would like to pick it up?

davidtazy:
did a proof of concept [https://github.com/davidtazy/conan-kcov](here) 

made 2 little patches from the original cmakelist : [https://github.com/davidtazy/kcov/commits/master](url) 

@SimonKagstrom  could you check them and tell if  its a conan build system problem or if the kcov CmakeLists could be "enhanced" ? 

if its a conan problem, the conan recipe can patch cmakelist.

SimonKagstrom:
I understand the libssl fix (thanks!), but I don't quite get the SOLIB patch and has commented it in the commit.

Doesn't the conan package build work otherwise?

davidtazy:
Dont know why, but ${SOLIB} is generated in `${CMAKE_BINARY_DIR}/lib`, so it  is not compiling unless this patch.
maybe a better fix should  be to use "generator expression" $<TARGET_FILE:${SOLIB}>  (seem to be available in cmake 2.8.4)

FYI cmake command generated by conan  is:
[...]

PR #322, opened 2020-05-04
make kcov available in conan package manager

davidtazy:
make kcov available in conan package manager #321


SimonKagstrom:
Merged, thanks a lot!

Issue #323, opened 2020-05-05
bash engine does not output nor drain stdout with recent bash


dciabrin:
I want to check code coverage of bash scripts that act as agents in a cluster manager [1].
Those scripts are regularly called by the cluster manager, and sometimes return some
xml output under regular operation.

Unfortunately, since #249, when a bash script is being instrumented by kcov, its stdout
is ignored altogether (with recent bash, which seems to be what everyone uses these days).
So I cannot use kcov transparently anymore and gather coverage data.

Another side effect of #249 is that since nothing is draining stdout of the instrumented
process, that process could get stuck in output and never finish successfully. For instance,
kcov currently hangs on this bash script:

    #!/bin/sh
    dd if=/dev/urandom of=/dev/stdout bs=65536 count=2 2>/dev/null | base64 -w0

[1] https://github.com/ClusterLabs/resource-agents/ for the record

SimonKagstrom:
Thanks a lot for debugging and fixing this!

Closing since your  pull request has been merged.

PR #324, opened 2020-05-05
Issue #323: bash engine: drain stdout of instrumented script

dciabrin:
When bash supports xtrace fd, no kcov marker can come from the
instrumented script's stdout, so it's safe to drain it as soon
as data is written to it, without waiting for '\n'.

stdout writers: Sum code/executed lines before generating headers


SimonKagstrom:
Thanks a lot for the fix!

Issue #331, opened 2020-07-15
Kcov reports 0% on bash scripts when running as root.


Orycterope:
Since Bash 4.4, you cannot set PS4 in the environment when running as root, due to security issues:

> 56. Bash-4.4 does not inherit $PS4 from the environment if running as root.

https://github.com/bminor/bash/blob/d233b485e83c3a784b803fb894280773f16f2deb/COMPAT#L383

But this is the way kcov does it, setting PS4 in the environment and then calling `/bin/bash -x <desired script>`. Which means kcov does not work when running as root. PS4 is not taken from the environment, and defaults to '+'. From there, kcov sees no PS4 line starting with `kcov@...` and reports that no line has been covered, hence the 0%. 

This is an important bug, because it affects the docker image that you provide, since in it kcov is run as root. Had me believe kcov did not work on bash at first, until I compiled it myself and tested it as an unprivileged user. 

What I propose to do is to set the PS4 in the `bash-helper.sh` that is sourced at the beginning instead, and stop setting it in the environment. 

I'm creating a PR which does just that.

SimonKagstrom:
OK, that sounds like a reasonable solution.

I actually believe the debug trap mode works as root though, but that has some other issues which makes it unsuitable.

PR #332, opened 2020-07-15
Bash when root: set PS4 in bash-helper instead of the environment.

Orycterope:
See #331 

The first commit is just there to prevent the line `PS4=...` in the helper to be printed when executed, which is inconvenient. There was no need to source the helper with -x, provided it sets -x anyway.

### DIFF
--- a/src/engines/bash-engine.cc
+++ b/src/engines/bash-engine.cc
@@ -221,12 +221,10 @@ public:
 
 			// Make a copy of the vector, now with "bash -x" first
 			char **vec;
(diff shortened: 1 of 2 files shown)


Orycterope:
Hummm ... I don't get how removing lines can decrease the coverage percentage 🤔

I hope this is not important for this PR to be merged.

SimonKagstrom:
The coverage decrease sounds like a bug in coveralls/codecov, so don't worry about that :-)

Anyway, good stuff and a nice solution, thanks a lot!

PR #333, opened 2020-07-27
Add regex to --replace-src-path

cattanimarco:
This change allows to pass a regex to the --replace-src-path argument. This is useful, for example, when the build directory of the tested binary is a randomly created temp directory. Note that after the change the older usage of the parameter is still valid

Usage example:
kcov --replace-src-path="([^ ]*)/build_dir/src":"/my_src_dir/"

### DIFF
--- a/src/filter.cc
+++ b/src/filter.cc
@@ -6,6 +6,7 @@
 #include <stdlib.h>
 #include <string>
 #include <vector>
+#include <regex>
 
 using namespace kcov;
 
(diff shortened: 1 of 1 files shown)


SimonKagstrom:
Good stuff, I can see why it can be useful!

However, could you also update the documentation in your changeset?

cattanimarco:
I will try to add the feature to the documentation too

PR #334, opened 2020-08-24
Prebuilt releases

l3laze:
Add automated build + release for tags via GitHub Actions. Builds for Linux on x86_64 natively, and uses Docker/chroot to build for x86, armv7, aarch64, and ppc64le.


SimonKagstrom:
Great stuff, thanks again!

PR #352, opened 2021-02-02
add libssl to Dockerfile system deps

brock-ff:


SimonKagstrom:
Thanks!

The CI build failed for some completely unrelated reason, an EULA was not accepted during package installation :-)

Issue #314, opened 2019-12-17
kcov returns incorrect exit code on abort signal


matdoering:
When kcov is called on an executable that terminates with signal 6, the error is not propagated to the return code of kcov. Instead, kcov simply returns an RC of 0.

Consider the following example (I wasn't able to construct a minimal example, so I'm just showing the debug 15 output from running kcov on a google test suite that produces the error):

> [ FATAL ] /tmp/googletest-release-1.8.0/googletest/src/gtest-port.cc:1020:: Only one stdout capturer can exist at a time.
> PT stopped PID 130 0x0000067f
> PT signal 6 at 0x7fffee28b336 for 130
> PT continuing 130 with signal 6
> PT stopped PID 157 0x00000000
> PT exit 0 at 0xffffffffffffffff for 157
> PT continuing 157 with signal 0
> PT error for 157: -1
> PT stopped PID 130 0x00000086
> PT terminating signal 6 at 0x7fffee28b336 for 130
> PT continuing 130 with signal 0
> PT error for 130: -1
> Returning error

After that 

> echo $? 

returns a value of 0. This is interesting because kcov seems to recognize the error and even states "Returning error" in the debug output.

Why is fixing this a high priority? When kcov is integrated into a build pipeline running using continuous integration, it should not miss failing tests. In this case, kcov misses the failing test but without kcov, the failure is detected because the correct exit code is returned.

I just took a quick dive into the code and I think what may cause the issue is this part:

https://github.com/SimonKagstrom/kcov/blob/afa0acea5ba9c7f14cd92a687d212ece28340a02/src/engines/ptrace.cc##L295-L296

From what I see, _ev.data_ has the correct error code set from _waitEvent()_. However, the next line in the code checks for  _ev_signal_ and only in this case uses ev.data and otherwise 0. I think that _ev.data_ should also be used for other signals than ev_signal (e.g. including sigabort). What do you think?

SimonKagstrom:
Thanks for the detailed and very good analysis!

Actually, I think the problem might be caused by the combination of your find and here:

https://github.com/SimonKagstrom/kcov/blob/master/src/collector.cc#L118

which only does an exit if the event type is ev_signal_exit. That in turn is perhaps caused by

https://github.com/SimonKagstrom/kcov/blob/master/src/engines/ptrace.cc#L249

i.e., if there are more threads/forked processes left it doesn't set ev_signal_exit. It should probably do that on e.g., the abort signal.

[...]

PR #365, opened 2021-09-30
#314 ev_signal_exit returns m_exitCode

dsteeley:
The exit code `m_exitCode` isn't set when crashing out with an `ev_signal_exit` so we don't pass the raised signal up.
By adding the same as the exit case, `m_exitCode = ev.data;` when a SEGFAULT is hit in the executed code kcov will return it to the caller.

This should resolve issue https://github.com/SimonKagstrom/kcov/issues/314

### DIFF
--- a/src/collector.cc
+++ b/src/collector.cc
@@ -135,7 +135,7 @@ private:
 						"kcov: binutils-dev), so the --verify option will not do anything."
 #endif
 						);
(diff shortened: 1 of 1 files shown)


SimonKagstrom:
Awfully sorry about the long delay for this one!

PR #366, opened 2021-09-30
Allow a single address to map to multiple lines

dsteeley:
kcov currently only allows an address to be mapped to a single line.

Rust produces coverage information of the form:

```
[root@9cc0f19f33ac default_ignore]# readelf --debug-dump=decodedline target/debug/default_ignore-addf76c983f37444 | grep -A 20 "CU: src/lib.rs"
CU: src/lib.rs:
File name                            Line number    Starting address
lib.rs                                         8             0x1e480
lib.rs                                        10             0x1e480
```

kcov reads this information for `lib.rs:8`, and gets the mapping for `0x1e480` and creates a mapping and life is good.

It then reads the information for `lib.rs:10` and gets the mapping for `0x1e480`, and overwrites the mapping for `lib.rs:8` (leaking it in the process), and life is now not good.

To resolve this, I've tweaked kcov, so that it's address mapping is now a `std::map<uint64_t, std::vector<Line*>>` instead of just being a `std::map<uint64_t, Line*>`.

### DIFF
--- a/src/reporter.cc
+++ b/src/reporter.cc
@@ -409,7 +409,7 @@ private:
 
 		for (LineList_t::iterator lit = it->second.begin();
                             lit != it->second.end();
-                            ++it) {
+                            ++lit) {
                         Line* line = *lit;
 
 			line->registerHit(addr, hits, m_maxPossibleHits != IFileParser::HITS_UNLIMITED);
--- a/src/reporter.cc
+++ b/src/reporter.cc
@@ -340,7 +340,7 @@ private:
 		uint64_t lineId = line->lineId();
 
 		line->addAddress(addr);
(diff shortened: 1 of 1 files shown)


SimonKagstrom:
Again, awfully sorry for letting this "rust" for over a month. Thanks a lot for your very nice debugging, investigation and implementation!

SimonKagstrom:
I'm investigating a problem with the bash tests caused by these commits, but I'll fix that issue.

SimonKagstrom:
The bash issue has been fixed with 652409e, but it would be good to get confirmation that it doesn't break the intended fix for Rust (and maybe other targets?).

PR #383, opened 2022-09-14
Fix build with binutils 2.39

oreo639:


SimonKagstrom:
Good stuff, thanks a lot!

PR #384, opened 2022-09-15
Fix build on gcc-13: add missing <stdint.h> include

trofi:
[ 15%] Building CXX object src/CMakeFiles/kcov.dir/writers/cobertura-writer.cc.o In file included from kcov/src/writers/cobertura-writer.cc:6: kcov/src/include/reporter.hh:24:90: error: 'uint64_t' has not been declared
   24 |      LineExecutionCount(unsigned int hits, unsigned int possibleHits, uint64_t order) :
      |                                                                       ^~~~~~~~

### DIFF
--- a/src/include/collector.hh
+++ b/src/include/collector.hh
@@ -2,6 +2,8 @@
 
 #include <string>
 
+#include <stdint.h>
+
 namespace kcov
 {
 	class IFileParser;
--- a/src/include/reporter.hh
+++ b/src/include/reporter.hh
@@ -3,6 +3,7 @@
 #include <string>
 
 #include <stddef.h>
+#include <stdint.h>
 
 namespace kcov
 {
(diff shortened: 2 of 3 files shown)


SimonKagstrom:
Great, thanks!

glebfm:
I believe the `<cstdint>` header would be more appropriate for C++ code.

SimonKagstrom:
Yes, you are right @glebfm. However, kcov would also be good to port to a modern C++ variant. It's currently C++03, but at least C++17 should be available on basically all systems which kcov runs on.

trofi:
One minor note: if you plan to switch to from `<stdint.h>` to `<cstdint>` you might need to make sure you are using `std::unit64_t` (instead of `unit64_t`) as standard only guarantees to provide them out of `std::` namespace.

PR #385, opened 2022-09-27
Add basic support for RISC-V architecture

r-value:
`riscv_EPC` value reference:
https://github.com/torvalds/linux/blob/a1375562c0a87f0fa2eaf3e8ce15824696d4170a/arch/riscv/include/asm/ptrace.h#L16


SimonKagstrom:
Amazing, thanks a lot!

PR #390, opened 2022-12-26
correct the license for debian folder

chenrui333:
Assuming the license declared in debian folder should be the source of truth.

https://github.com/SimonKagstrom/kcov/blob/79a0e1f67a0c8a0c2c12ef2c5a9296ee9be2492c/debian/copyright#L5-L8

Then this PR is needed to correct the github license reporting (relates to https://github.com/Homebrew/homebrew-core/pull/119013)

### DIFF
--- a/debian/copyright
+++ b/debian/copyright
@@ -5,16 +5,16 @@ Source: <url://example.com>
 Files: *
 Copyright: <years> <put author's name and email here>
            <years> <likewise for another author>
-License: GPL-3.0+
+License: GPL-2.0+
 
 Files: debian/*
 Copyright: 2018 Paul Dreik <github@pauldreik.se>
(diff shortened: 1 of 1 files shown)


SimonKagstrom:
Sorry, the license is GPLv2, so the debian/copyright is the incorrect one.

chenrui333:
> Sorry, the license is GPLv2, so the debian/copyright is the incorrect one.

ok, let me change the debian folder then.

chenrui333:
@SimonKagstrom let me know if that works for you now. Thanks!

SimonKagstrom:
Thanks a lot, well spotted!

chenrui333:
Thanks @SimonKagstrom!

Issue #397, opened 2023-06-15
Broken symbolic link


Hanaasagi:
When the out-directory is a relative path, it appears that the generated symbolic links are incorrect.

![2023-06-15_01-34](https://github.com/SimonKagstrom/kcov/assets/9482395/3d0eea49-f92a-4dcc-abf0-a9bb246a163a)

PR #398, opened 2023-06-15
Convert to an absolute path before creating symbolic link

Hanaasagi:
Fix https://github.com/SimonKagstrom/kcov/issues/397

- Get the absolute path of `out-directory` using `realpath`.

### DIFF
--- a/src/output-handler.cc
+++ b/src/output-handler.cc
@@ -41,11 +41,11 @@ namespace kcov
 			{
 				(void)unlink(readableName.c_str());
 
-                std::string out_path = get_real_path(m_outDirectory);
-                if (symlink(out_path.c_str(), readableName.c_str()) < 0)
-                {
-                    kcov_debug(INFO_MSG, "Can't symlink readable name\n");
-                }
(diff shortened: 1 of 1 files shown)


src/output-handler.cc
```diff
@@ -40,10 +42,18 @@ namespace kcov
 			if (!conf.keyAsInt("cobertura-only"))
 			{
 				(void)unlink(readableName.c_str());
-				if (symlink(m_outDirectory.c_str(), readableName.c_str()) < 0)
-				{
-					kcov_debug(INFO_MSG, "Can't symlink readable name\n");
-				}
+
+                char buf[MAXPATHLEN + 1];
+                char *out_path = realpath(m_outDirectory.c_str(), buf);
```
SimonKagstrom, line 47: Nice! An improvement though: there is a get_real_path in utils.hh, which outputs a std::string

Hanaasagi: done.

SimonKagstrom:
Merged, thanks!

PR #399, opened 2023-06-30
Add basic support for LoongArch architecture

FreeFlyingSheep:
LoongArch documentation: https://loongson.github.io/LoongArch-Documentation/README-EN.html
ptrace reference: https://github.com/torvalds/linux/blob/82a2a51055895f419a3aaba15ffad419063191f0/arch/loongarch/include/uapi/asm/ptrace.h#L25)
breakpoint reference: https://github.com/torvalds/linux/blob/82a2a51055895f419a3aaba15ffad419063191f0/arch/loongarch/include/uapi/asm/break.h#L12


SimonKagstrom:
Beautiful, thanks a lot! Also nice with the references.

Issue #405, opened 2023-08-22
Testing the new mach-engine with Go


myzhan:
```bash
$ mkdir build && cd build
$ cmake -G Xcode -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl -DOPENSSL_LIBRARIES=/usr/local/opt/openssl/lib ..
$ xcodebuild -target kcov -configuration Release
note: Run script build phase 'CMake PostBuild Rules' will be run during every build because the option to run the script phase "Based on dependency analysis" is unchecked. (in target 'kcov' from project 'kcov')
PhaseScriptExecution CMake\ PostBuild\ Rules /Users/zhanqp/github/kcov/build/build/kcov.build/Release/Script-8BEB10D058631033A81B2400.sh (in target 'kcov' from project 'kcov')
    cd /Users/zhanqp/github/kcov
    /bin/sh -c /Users/zhanqp/github/kcov/build/build/kcov.build/Release/Script-8BEB10D058631033A81B2400.sh
Signing the kcov binary with an ad-hoc identity
kcov: No such file or directory
Command PhaseScriptExecution failed with a nonzero exit code

note: Run script build phase 'Generate CMakeFiles/ZERO_CHECK' will be run during every build because the option to run the script phase "Based on dependency analysis" is unchecked. (in target 'ZERO_CHECK' from project 'kcov')
** BUILD FAILED **


The following build commands failed:
	PhaseScriptExecution CMake\ PostBuild\ Rules /Users/zhanqp/github/kcov/build/build/kcov.build/Release/Script-8BEB10D058631033A81B2400.sh (in target 'kcov' from project 'kcov')
(1 failure)
```

After manually signing the binary, It failed to collect coverage from a go binary.

$ ./src/Release/kcov $HOME/tmp/report $HOME/tmp/hello
Picked parser: Mach-O
warning: no debug symbols in executable (-arch x86_64)

SimonKagstrom:
If there are no debug symbols, it will not work. Normally, you pass -g to the compiler, but I'm not sure how go does it?

The other issue seems to be with xcode. Perhaps it would be better to build it with make or ninja, i.e.,

```cmake -G make -DOPENSS...```

Does that work?

myzhan:
For the building issue, I switch to make and it's ok. And I think the version of dsymutil does not support go executable, which I installed via brew. Do we need a new go-parser for go executables?

SimonKagstrom:
[...]

PR #408, opened 2023-10-10
Work with Go binary on Mac

myzhan:
See: https://github.com/SimonKagstrom/kcov/issues/405

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@
 kcov
 build
 obj-*/
+.cache
--- a/src/configuration.cc
+++ b/src/configuration.cc
@@ -642,6 +642,7 @@ public:
 		setKey("cobertura-full-paths", 0);
 		setKey("codecov-full-paths", 0);
 		setKey("cobertura-only", 0);
+		setKey("is-go-binary", 0);
 	}
 
 	void setKey(const std::string &key, const std::string &val)
--- a/src/engines/mach-engine.cc
+++ b/src/engines/mach-engine.cc
@@ -16,6 +16,7 @@
 #include <engine.hh>
 #include <mach-o/loader.h>
 #include <mach/mach.h>
+#include <mach/mach_error.h>
 #include <mach/mach_init.h>
 #include <mach/mach_types.h>
 #include <mach/port.h>
@@ -35,6 +36,7 @@
 #include <unistd.h>
 #include <unordered_map>
 #include <utils.hh>
+#include <iostream>
 
 using namespace kcov;
 extern char** environ;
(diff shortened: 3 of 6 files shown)


SimonKagstrom:
is-go-binary is clear enough I think!

myzhan:
> is-go-binary is clear enough I think!

Because the whole file content need to be read into memory, so I think we can use a commandline option to set "is-go-binary" to true directly, then we don't need to read the whole file.

SimonKagstrom:
Yes. You can always do `--configure=is-go-binary=1` to just set the key, but otherwise a command line option can also be added, at least for OSX.

SimonKagstrom:
> Oops, don't know why some tests failed.

The FreeBSD problem is some infrastructure issue, but the OSX tests should run. Do non-go programs work with your patch?

myzhan:
OK, I fixed a segfault when running non-go programs. 

But I found a new issue. Tried to generate dSYM by clang with -g, but the dsymutil modified it even the file existed, and then kcovfailed to read dwarf info. After commented out the dysmutil part, everything worked.

How to reproduce.

```bash
$ clang -v
Apple clang version 13.1.6(clang-1316.0.21.2.5)
$ dsymutil -v
Apple LLVM version 13.1.6(clang-1316.0.21.2.5)
$ clang -g hello.c -o hello
$ ll hello.dSYM/Contents/Resource/DWARF/hello
8.6k size
$ kcov ./report ./hello
warning: (x86_64) /my/tmp/folder/on/the/mac/hello-3838c7.o unable to open object file: No such file or directory
warning: no debug symbols in executable (-arch x86_64)
hello
$ ll hello.dSYM/Contents/Resource/DWARF/hello
8.2k size

**Comment by SimonKagstrom** [comment]
I believe clang puts DWARF info directly into the executable if it's compiled in a single step, i.e., without linking. This is a case I've ignored, since it basically only affects toy programs. I guess it still works fine when running with linked binaries?

**Comment by myzhan** [comment]
> I believe clang puts DWARF info directly into the executable if it's compiled in a single step, i.e., without linking. This is a case I've ignored, since it basically only affects toy programs. I guess it still works fine when running with linked binaries?

Clang puts DWARF info into the dSYM file, I can check with objdump.

```bash
$ objdump --macho -x ./hello
./hello:

Sections:
Idx Name          Size     VMA              Type
  0 __text        0000000f 0000000100003fa0 TEXT
  1 __unwind_info 00000048 0000000100003fb0 DATA
```

```bash
$ objdump --macho -x hello.dSYM/Contents/Resources/DWARF/hello
...
Section
  sectname __debug_info
   segname __DWARF
      addr 0x000000010000506e
      size 0x0000000000000053
    offset 8302
     align 2^0 (1)
    reloff 0
    nreloc 0
      type S_REGULAR
attributes (none)
 reserved1 0
 reserved2 0
...
```

SimonKagstrom:
OK, I see! So what really happens is that clang generates the dSYM info directly then? I guess this is where it would be good to be able to check that the dSYM and binary matches, and ignore dsymutil in this case. I'm on Linux right now, so can't check this locally.

Anyway, this is unrelated to your change, so merging that!

myzhan:
> OK, I see! So what really happens is that clang generates the dSYM info directly then? I guess this is where it would be good to be able to check that the dSYM and binary matches, and ignore dsymutil in this case. I'm on Linux right now, so can't check this locally.
> 
> Anyway, this is unrelated to your change, so merging that!

Yes, clang generates the dSYM info directly.

PR #410, opened 2023-11-01
fix dead link in output

luehmann:


SimonKagstrom:
Hehe! Thanks for the fix!

PR #416, opened 2024-03-08
configuration: refactorize code scanning for executables

perillo:
Add the `look_path` function for searching executable in directories named by the `PATH` environment variable in the `utils` source file.

Refactorize the `parse` function in configuration.cc, using the new `look_path` function.  This greatly simplify the code.

This PR implements an idea discussed #415.

The new change **should** behave the same as the current implementation, but unfortunately I can not verify this, because many tests fails on my system, probably due to incorrectly building the test suite.

### DIFF
--- a/src/include/utils.hh
+++ b/src/include/utils.hh
@@ -223,3 +223,7 @@ static inline uint32_t hash_block(const void *buf, size_t len)
 }
 
 uint32_t hash_file(const std::string &filename);
+
+// Searches for an executable named file in the directories named by the PATH
+// environment variable.
+int look_path(const std::string &file, std::string *out);
--- a/src/utils.cc
+++ b/src/utils.cc
@@ -775,3 +775,36 @@ uint32_t hash_file(const std::string &filename)
 
 	return out;
 }
+
+int look_path(const std::string &file, std::string *out)
+{
+	const char *sys_path = getenv("PATH");
+	if (!sys_path)
+		return -1;
+
+	const std::vector<std::string> sys_paths = split_string(sys_path, ":");
+	for (std::vector<std::string>::const_iterator it = sys_paths.begin();
+			it != sys_paths.end(); ++it)
+	{
+		const std::string root = *it;
+		const std::string path = get_real_path(root + "/" + file);
+		struct stat st;
+
+		if (lstat(path.c_str(), &st) < 0)
+			continue;
+
+		// Regular file?
+		if (S_ISREG(st.st_mode) == 0)
+			continue;
+
+		// Executable?
+		if ((st.st_mode & (S_IXUSR | S_IXGRP | S_IXOTH)) == 0)
+			continue;
+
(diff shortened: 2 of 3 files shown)


src/utils.cc
```diff
@@ -775,3 +775,36 @@ uint32_t hash_file(const std::string &filename)
...
 	return out;
 }
+
+int look_path(const std::string &file, std::string *out)
+{
+    const char *sys_path = getenv("PATH");
+	if (!sys_path)
+		return -1;
+
+    const std::vector<std::string> sys_paths = split_string(sys_path, ":");
```
SimonKagstrom, line 785: This looks like a mix of tabs and space, or otherwise the indentation is strange?

src/include/utils.hh
```diff
@@ -223,3 +223,7 @@ static inline uint32_t hash_block(const void *buf, size_t len)
 }
 
 uint32_t hash_file(const std::string &filename);
+
+// Searches for an executable named file in the directories named by the PATH
+// environment variable.
+int look_path(const std::string &file, std::string *out);
```
SimonKagstrom, line 229: You don't have to do that here and now, but I think it's high time to ditch c++03 in kcov (and the MacOS port already depends on newer), and then this would be better to write as

`std::optional<std::string> look_path(...)`

src/utils.cc
perillo: I configured my editor to use space for indentation.
When I found that `kcov` code use tabs, I added an `.editorconfig` file with the correct style, but since I updated the code manually some line where not formatted correctly.

It seems that the same problem is also in other functions.

src/include/utils.hh
perillo: Looking at CMake, for building `kcov` you use `--std=c++17`, for tests `--std=c++0x` and, only once, `--std=c++1y` (that is deprecated after C++14).

perillo:
Is seems that this change is also incorrect: https://github.com/SimonKagstrom/kcov/actions/runs/8208545224.

I was sure that the behavior was the same as the original code.

src/include/utils.hh
SimonKagstrom: Yes, the C++1y was used for the unit tests, which are sort of stale now anyway. I moved to c++17 after som MacOS rewrites, but forgot that it was done everywhere and not just for the Mac build (which is probably all code that needs it).

Anyway, you don't have to make that change here, I think it's fine as it is.

src/utils.cc
SimonKagstrom: Right. I've started outs with tab indentation, but I've also later come to prefer spaces, so the code base is probably mixed.

I guess I should reformat it all sometime in the future...

SimonKagstrom:
Let's merge this anyway, I'll resolve the test errors later. Thanks a lot!

perillo:
Thanks to you.

PR #417, opened 2024-03-09
python-helper: update deprecated code

perillo:
Currently `python-helper.py` imports the `imp` module, but this module has been deprecated since version 3.4 and removed in version 3.12.

Refactorize the code to handle Python versioning by moving the code at the start of the file.

Use one block for python>=3.0, declaring the report_trace_real function. Use a second block for python>=3.5, declaring the new_module function.

Add tests for python2 support.

Fix incorrect code using comparison to `None` instead of `cond is None`. Reported by ruff check.

Format the code with ruff format, with line-length = 100.

### DIFF
--- a/src/engines/python-helper.py
+++ b/src/engines/python-helper.py
@@ -4,33 +4,52 @@
 # Based on http://pymotw.com/2/sys/tracing.html, "Tracing a program as it runs"
 # and http://hg.python.org/cpython/file/2.7/Lib/trace.py
 
-import imp
 import sys
 import os
 import struct
 
 fifo_file = None
-report_trace_real = None
 
-try:
-    # In Py 2.x, the builtins were in __builtin__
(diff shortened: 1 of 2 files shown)


perillo:
Before this change, when building `kcov` from old master, I got a lot of errors, but now this is fixed.  I'm not sure what was causing these failures, probably the deprecation warning for the `imp` module, when using python3.11.

The only tests that have problems are:
  - `bash.bash_stderr_redirection.runTest`
    I suspect the problem is that `/usr/bin` is dash and not bash on my system.
    The fix is simple and I plan to create a new PR.

  - `system_mode.system_mode_can_record_and_report_binary.runTest`
    No idea about why it hangs forever.

SimonKagstrom:
The system mode test can be removed. System mode has never worked satisfactory anyway, so can be disabled until some major future refactoring.

Thanks a lot for the PR!

PR #421, opened 2024-03-11
ci.yml: update vmactions/freebsd to v1

perillo:
v1 is the most recommended to use and runs on ubuntu-latest instead of macos-12.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,13 +60,13 @@ jobs:
 
 
   build-freebsd:
-    runs-on: macos-12
+    runs-on: ubuntu-latest
     name: Build and test FreeBSD executable
     steps:
     - uses: actions/checkout@v3
     - name: FreeBSD
       id: test
-      uses: vmactions/freebsd-vm@v0
+      uses: vmactions/freebsd-vm@v1
       with:
         usesh: true
         sync: rsync


SimonKagstrom:
Thanks!

Yes, I have another branch where I'm trying to get this working: https://github.com/SimonKagstrom/kcov/tree/fix-freebsd-python2-tests, which also uses v1 now.

However, for some reason, the bash tests no longer works with this one...

perillo:
> Thanks!
> 
> Yes, I have another branch where I'm trying to get this working: https://github.com/SimonKagstrom/kcov/tree/fix-freebsd-python2-tests, which also uses v1 now.
> 
> However, for some reason, the bash tests no longer works with this one...

I suspected that you also have patches ready.  But why not `ubuntu-latest` ?

perillo:
> However, for some reason, the bash tests no longer works with this one...

I suspect the reason is that the default shell in FreeBSD may be tcsh.  So `/bin/sh` can use tcsh instead of a POSIX shell.

SimonKagstrom:
It turned out to be a problem with file descriptor numbers. I fixed it with e27cb33e5b25b48689eda1178133e7ec90f20c43 in the PR, which basically just disables it for FreeBSD. I don't think it's a very big issue.

Probably broke with the newer FreeBSD I guess.

As for ubuntu-latest, I'm sort of worried that the FreeBSD build will break again when 24.04 is released soonish. :-)

SimonKagstrom:
Anyway, I'll close this PR since pretty much the same change is also part of #422, but thanks a lot for the discussion and insights!

Issue #423, opened 2024-03-13
Merge some/all of the zig improvements by liyu1981


SimonKagstrom:
I.e., this fork:

https://github.com/liyu1981/kcov

@liyu1981, some input?

liyu1981:
@SimonKagstrom I am glad that you find this work.

Currently the implementation of my hack is based on string comparison as from the semantics level I do not find good ways of doing that. I am not sure whether this is something `kcov` prefer, but I can prepare it in a pull request for discussion.

SimonKagstrom:
I think the string comparison is fine. I guess probably other tools would do it that way as well.

The only real comment I have (see the PR) is that I think the comparison can be done in reporter.cc instead, and not in each writer, but feel free to try that out!

Thanks!

PasVegan:
any update ? I see that the PR isn't merged.

SimonKagstrom:
I think it's up to @liyu1981 to investigate and maybe fix what's left

mnemnion:
Heavy Zig + kcov user here, dropping by the issue board trying to figure out fugitive thread stuff. 

Just to say, I build a kcov step like this:

```zig
    const run_kcov = b.addSystemCommand(&.{
        "kcov",
        "--clean",
        "--exclude-line=unreachable,expect(false),panic(,no-coverage",
    });
    run_kcov.setCwd(b.path("."));
    run_kcov.addPrefixedDirectoryArg("--include-pattern=", b.path("src"));
    const coverage_output = run_kcov.addOutputDirectoryArg(".");
    run_kcov.addArtifactArg(module_unit_tests);

    run_kcov.enableTestRunnerMode();

    const install_coverage = b.addInstallDirectory(.{
        .source_dir = coverage_output,
        .install_dir = .{ .custom = "coverage" },
        .install_subdir = "",
    });

    const coverage_step = b.step("coverage", "Generate coverage (kcov must be installed)");
    coverage_step.dependOn(&install_coverage.step);
```
While it would be nice, I suppose, to 'lift' the `--exclude-line` into kcov itself, I would not personally notice if it happened.  

This just seemed like a good place to say thanks, Simon, for your excellent work, which has spared me many hassles down the road! That and leave a little Zig kcov recipe where others might find it.

[...]

PR #424, opened 2024-03-13
Kcov zig 4merge (for discussion)

liyu1981:
#423 
a preview of core changes to kcov for discussion


README.md
```diff
@@ -1,13 +1,62 @@
+# A fork of `kcov` for better covering Zig
+
+This is a rather than naive fork of wonderful `kcov` for better supporting Zig projects.
+
+## It adds the ability to auto recognize Zig's `unreachable` and `@panic` for auto ignoring.
+
+Here is a screenshot of auto ignoring `unreachable`.
+
+![zig-unreachable-example](https://github.com/liyu1981/kcov/blob/master/nocover/zig-unreachable.png?raw=true)
```
SimonKagstrom, line 9: All this is nice information, but when "native" kcov supports it, I think it won't be needed (as it's shown in the output anyway).

README.md
```diff
@@ -1,13 +1,62 @@
...
+cd kcov
+mkdir build
+cd build
+CC="clang" CXX="clang++" cmake -G Ninja ..
+ninja
+```
+
+after building is done. The binary is at `build/src/kcov`. Copy somewhere and use it.
+
+(Or you can download a copy of Apple Silicon version binary from the release section.)
```
SimonKagstrom, line 40: I think we can make a release shortly after merge of this, and then the osx homebrew is usually updated quite quickly.

I suppose that can be added to the README, or another documentation file. Since not all distributions are up to date, I don't want to in general recommend installing from there.

src/writers/codecov-writer.cc
```diff
@@ -134,6 +135,12 @@ class CodecovWriter : public WriterBase
 				IReporter::LineExecutionCount cnt = m_reporter.getLineExecutionCount(file->m_name, n);
 				std::string hitScore = "0";
 
+				std::string& line_str = file->m_lineMap.at(n);
```
SimonKagstrom, line 138: Kcov now builds with c++17, so you can use auto & for these.

src/writers/html-writer.cc
```diff
@@ -252,15 +261,29 @@ class HtmlWriter : public WriterBase
...
+			}
+
 			// Skip entries (merged ones) that shouldn't be included in the totals
 			if (summary.m_includeInTotals)
 			{
-				nTotalCodeLines += summary.m_lines;
-				nTotalExecutedLines += summary.m_executedLines;
+					nTotalCodeLines += inFilesTotalCodeLines;
+					nTotalExecutedLines += inFilesTotalExecutedLines;
+					// nTotalCodeLines += summary.m_lines;
```
SimonKagstrom, line 279: Remove the comment

src/writers/html-writer.cc
```diff
@@ -252,15 +261,29 @@ class HtmlWriter : public WriterBase
...
+					// nTotalExecutedLines += summary.m_executedLines;
 			}
 
-			std::string datum = getIndexHeader(fmt("%s/index.html", de->d_name), name, name, summary.m_lines,
-					summary.m_executedLines);
+			std::string datum = getIndexHeader(fmt("%s/index.html", de->d_name), name, name, inFilesTotalCodeLines,
+					inFilesTotalExecutedLines);
+			// std::string datum = getIndexHeader(fmt("%s/index.html", de->d_name), name, name, summary.m_lines,
```
SimonKagstrom, line 285: ... also here

src/writers/json-writer.cc
```diff
@@ -61,12 +62,21 @@ class JsonWriter : public WriterBase
 			for (unsigned int n = 1; n < file->m_lastLineNr; n++)
 			{
 				IReporter::LineExecutionCount cnt = m_reporter.getLineExecutionCount(file->m_name, n);
+
+				const std::string& line_str = file->m_lineMap[n];
+				const std::string& file_name = it->first;
+				bool should_cover = shouldCover(line_str, file_name);
```
SimonKagstrom, line 68: I think that all these can probably be moved to reporter.cc instead, to the `lineIsCode` method. I.e., you could do something like `shouldCover(...) && it->second->lineIsCode(...)` there, and then you can probably remove all changes to the writers.

At least it's something to try out.

src/writers/nocover.cc
```diff
@@ -0,0 +1,84 @@
...
+bool endsWith(const std::string& full_string, const std::string& ending) {
+    if (full_string.length() >= ending.length()) {
+        return (full_string.compare(full_string.length() - ending.length(), ending.length(), ending) == 0);
+    } else {
+        return false;
+    }
+}
+
+FileType getFileExt(const std::string& file_name) {
+  const std::string zig_ending = ".zig";
```
SimonKagstrom, line 24: here you could also write this as

```
constexpr auto relevant_extensions = std::array {
   std::pair{".c", c},
   [...]
};

for (const auto &[ending, extension])
{
    if (endsWith(..., ending)
    {
       return extension;
    }
}
```

README.md
liyu1981: agree, I will remove this in later cleaned up PR.

README.md
liyu1981: same to above

src/writers/codecov-writer.cc
liyu1981: got it.

src/writers/html-writer.cc
liyu1981: will do

src/writers/json-writer.cc
liyu1981: good suggestion, will do

paulofduarte:
I looked into this issue for a few days and identified a very good solution, but not sure very upstreamable, at least not without some refactoring in kcov.

Zig uses DWARF5 that cause issues with libdw/libdarwin, also I never liked the fact that for running it macOS installing libdarwin from homebrew. Even working around this Zig generates a lot of unreachable code that causes an artificial drop in coverage.

Zig has it's own DWARF library, so I create 2 shims to replace `parsers/dwarf.cc` and `parsers/macho-parser.cc`. on the build. This fixed the DWARF5 and allow me to implement filtering of the unreachable code. macOS now build without installing any library. If you want to check the changes they are:

https://github.com/allyourcodebase/kcov/pull/2

SimonKagstrom:
Cool @paulofduarte !

I'm open for any fix, although I'm not a Zig programmer (yet?). There's also #496, which seems like it's related / the same as this?

paulofduarte:
Hi @SimonKagstrom , #496 is definitely related. I think a good approach would be to extract Zig's own lib for DWARF and Mach-O and use it for kcov, I am not sure how general the library is for non Zig, but will take a look on how feasible is this approach.

SimonKagstrom:
Great, thanks a lot @paulofduarte !

SimonKagstrom:
I see that https://github.com/roc-lang/zig-kcov might be a more recent version of this. Perhaps mergeable to "upstream" kcov?

PR #427, opened 2024-03-16
configuration: improve getopt_long error reporting

perillo:
Currently, in case of an unrecognized option we get two error messages. The first is printed by `getopt_long` and the second is printed by `kcov` error handler.

An additional issue is that the `kcov` error handler prints `optopt`, but this value is `0` (NULL character) in this case.

Improve the error handler by using `optind` to get the offending option. This is done in a separate switch case, instead of the default case. Disable `getopt` error messages by setting `opterr` to 0.

Configure `getopt_long` to report a missing argument using a different error code, so that this can be handled in a separate switch case. Note however that this error is only reported when the option is the last argument.

Update the `basic.TooFewArguments` and `basic.WrongArguments` test cases.

## TODO

Print useful error message in case of an invalid option argument.  This PR will not implement it.

### DIFF
--- a/src/configuration.cc
+++ b/src/configuration.cc
@@ -171,6 +171,7 @@ public:
 		/* Hooray for reentrancy... */
 		optind = 0;
 		optarg = 0;
+		opterr = 0;
 		while (1)
 		{
 			int option_index = 0;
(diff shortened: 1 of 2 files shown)


perillo:
The new test should probably be skipped, since it is preventing full testing.

SimonKagstrom:
That's a nice improvement of the option parsing, thanks a lot!

PR #428, opened 2024-03-17
tests: fix basic.OutDirectoryIsExecutable

perillo:
Commit 4b4cb9c (Merge pull request #426 from perillo/fix-argv-parsing) added a regression test, but the test failed due to `ptrace` errors on Linux.
The "Operation not permitted" error was caused by `kcov` trying to trace the "/bin/usr/python".

The solution is to execute an executable having user permission. Restore the `short-test.py` executable, and use it instead of `python`. `short-test.py` always exits with exit status 0.

### DIFF
--- a/tests/python/short-test.py
+++ b/tests/python/short-test.py
@@ -1,9 +1,9 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
 
 import sys
 
 def doSomething():
-    print 'Hello world'
+    print('Hello world')
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


perillo:
For the reason of the `ptrace` error with the old test, see
https://bugs.launchpad.net/ubuntu/+source/linux-signed/+bug/1860822

> In Linux, process tracing is tied with the process dumping flag. The dumpable flag is cleared at execve(2) time when a setuid or setgid application is executed. This flag persists to child processes created by fork(2) and will only be reset when a process calls execve(2) again.

```console
> stat -L /usr/bin/python3
Access: (0755/-rwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
```

```console
stat tests/python/short-test.py 
Access: (0750/-rwxr-x---)  Uid: ( 1000/  manlio)   Gid: ( 1000/  manlio)
```

The kcov test suite runs correctly on my system, without using sudo and with `/proc/sys/kernel/yama/ptrace_scope` set to `1` (the default value).

SimonKagstrom:
OK, thanks for the PR and the well-made analysis! Merged!

PR #431, opened 2024-03-26
Improve tests code more

perillo:
Additional improvements to `kcov` test suite.

  - [x] tests: validate `testbase.configure` arguments
    
    Ensure that all the paths in `testbase.configure` are not empty and
    normalized.
    
    Ensure that `outbase` is not the current working directory.

  - [x] tests: convert simple % style string format to f-string
    
    Convert old formatting style to f-string style, using
    `python3 -m libcst.tool codemod convert_percent_format_to_fstring.ConvertPercentFormatStringCommand`

  - [x] tests: remove unnecessary call to `setUp` method.
    
    It is not necessary because `setUp` is called automatically by
    `TestCase.run`.

  - [x] tests: use `skipTest` instead of print and return
    
    Using `skipTest` is the preferred method for skipping tests.
    Now the test message is printed on the same line.

  - [x] tests: override `TestCase.tearDown` in `testbase.py`
    
    Use `os.makedirs` to create the `kcov` directory in `setUp` and `shutil.rmtree`
    to remove it in `tearDown`.
    
    Don't use the shell, since errors are not propagated.

  - [x] Ensure the "kcov-kcov" directory uses a fixed path
    
    In `ci.yml`, `codecov` assumes that the output directory is
    "/tmp/kcov-kcov".

    Don't assume that `outbase` is always "/tmp", and instead use the
    "/tmp/kcov-kcov" path.

## NOTES

  * I used tools to automate refactoring, like `LibCST` and `sed`.

  * LibCTS codemod for converting old format to f-string unfortunately added parenthesis in some non simple statement.  Only later I found that there was another tool `flynt` but I have not used it to see if there are differences in formatting.

  * When removing calls to `setUp` method, I also inadvertently removed calls to `setUp` method inside the `doTest` method.  This is fine, since `doTest` is always called from the `runTest` method so `setUp` is always called.

## TODO

I noted that `doShell` is used only two times; other code using the shell use `system`.

Additionally, both `doShell` and `system` **does not** propagate errors, so I plan to make this more robust in a separate PR.  I also plan to group shell commands, instead of spawning a subprocess for every command line.

### DIFF
--- a/tests/tools/accumulate.py
+++ b/tests/tools/accumulate.py
@@ -1,4 +1,3 @@
-
 import parse_cobertura
 import testbase
 
@@ -200,9 +199,7 @@ class merge_coveralls(testbase.KcovTestCase):
             + "/tests/bash/shell-main"
         )
 
-        rv, o = self.doShell("grep second.py %s/kcov/main/coveralls.out" % (testbase.outbase))
+        rv, o = self.doShell(f"grep second.py {(testbase.outbase)}/kcov/main/coveralls.out")
         assert rv == 0
-        rv, o = self.doShell(
-            "grep shell-main %s/kcov/shell-main/coveralls.out" % (testbase.outbase)
-        )
+        rv, o = self.doShell(f"grep shell-main {(testbase.outbase)}/kcov/shell-main/coveralls.out")
         assert rv == 0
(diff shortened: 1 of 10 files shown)


SimonKagstrom:
Beautiful! Thanks a lot for taking on the testsuite!

perillo:
Thanks to you for the review.

PR #432, opened 2024-03-30
Improve tests code even more

perillo:
Additional improvements to kcov test suite.

## IMPORTANT

This PR is not ready to be merged.

## CHANGES

  - [x] Additional improvements to kcov test suite.
    The current, old API, is not only inconvenient (using a thread and a timer to implement timeouts) but it is also incorrect since the child process is not waited.
    
    Use subprocess.run for both `doShell` and `do`.  Note that for the `do` method, this is a behavior change, since a child process is always terminated by kill.
    
    Add a default timeout, set to 10 min.
    
    After this change the test `system_mode_can_record_and_report_binary` fails.  This will be addressed in the next commit.

### DIFF
--- a/tests/tools/bash_linux_only.py
+++ b/tests/tools/bash_linux_only.py
@@ -29,7 +29,6 @@ class bash_exit_before_child(testbase.KcovTestCase):
             + "/tests/bash/background-child.sh",
             kcovKcov=False,
             timeout=3.0,
-            kill=True,
         )
         self.assertEqual(0, rv, "kcov exited unsuccessfully")
         dom = parse_cobertura.parseFile(
--- a/tests/tools/testbase.py
+++ b/tests/tools/testbase.py
@@ -6,14 +6,17 @@ import platform
 import shutil
 import subprocess
 import sys
-import threading
 import unittest
+
+PIPE = subprocess.PIPE
 
 kcov = ""
 kcov_system_daemon = ""
 outbase = ""
 testbuild = ""
 sources = ""
+
+default_timeout = 10 * 60
 
 
 # Normalize path, also ensuring that it is not empty.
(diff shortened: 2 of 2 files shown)


perillo:
As I suspected `system_mode_can_record_and_report_binary` did not failed on Github CI.  Probably because the tests are run as root, while on my PC the tests are run as a normal user.

The strange thing is that `system_mode_can_record_and_report_binary` **is** incorrect, since it starts the daemon process but **does not** stop it.

SimonKagstrom:
Thanks!

As for system-mode, it has really never been properly working anyway. It's sort of meant to be run as root, since the purpose is to collect coverage from all binaries in a system, as a part of the startup of an embedded system. However, it really was never working outside of a proof-of-concept stage, so I think we can simply disable the system mode tests for now.

perillo:
Thanks for the explanation.

However it is true that before the change introducing `subprocess.run`, the test worked until the `skipTest` with normal privilege.

Do you have a preferred channel for communication, like email, forum or github issue?
For the following commits I may need your help for some clarification.

SimonKagstrom:
I'm fine with pr communication!

What I mean is mainly that we can just remove the system mode tests. Maybe also the build of the binary, since it's not really working other than as a toy.

SimonKagstrom:
Good stuff! The tests feel much more proper now, and not like the plain hacks that they were before :-)

Anyway, I've looked through your changes, and I think they look good! Tell me when you're ready and I'll merge them!

perillo:
I forgot to add a commit.

SimonKagstrom:
Big thanks!

In general, if I remember correctly, the kcovKcov stuff was to avoid running kcov on itself when collecting data for compiled code, where it's not possible to trace kcov itself since it would mean recursive ptrace, which is illegal. Thus, coverage data on kcov itself can be collected when covering bash and python, but not for compiled code.

PR #433, opened 2024-04-03
workflows: fix ci-run-tests.sh script

perillo:
The run function is incorrect, since it does not accept additional arguments.

Don't use `pwd` for sources directory, since all other paths, excluding /tmp, are relative.

Update the ci.yml workflow, removing the "amd64" argument to the ci-run-tests.sh script.  It was working because the argument was ignored.

@SimonKagstrom, the change from "`pwd`" to `"."` may be controversial.  Let me know what you think.  I use it in my `test.sh` script, when testing locally.

### DIFF
--- a/.github/workflows/ci-run-tests.sh
+++ b/.github/workflows/ci-run-tests.sh
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 run () {
-  tests/tools/run-tests build/src/kcov /tmp/ build-tests/ `pwd` -v || exit 64
+  tests/tools/run-tests build/src/kcov /tmp/ build-tests/ "." -v "$@" || exit 64
 }
 
 run "$@"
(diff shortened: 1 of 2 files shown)


SimonKagstrom:
I think that's fine, since this is anyway used in a controlled environment.

I personally just use the `run-tests` script directly, when running the tests.

perillo:
> I think that's fine, since this is anyway used in a controlled environment.
> 
> I personally just use the `run-tests` script directly, when running the tests.

I use a custom script: https://gist.github.com/perillo/cebd9f82094471b241b961a4f869b0ed

The script use `python -m libkcov`, the new way to run the tests (the PR will be pushed soon).

By the way: in the `cleanup` function I have listed all the generated files; is there a reason why `main.log`, `thread.log` and `.debug` are not written in `/tmp` ?

In a future PR I plan to move these files to `$TMPDIR` and later add cleanup functions to remove them.  The purpose it enabling support for looping the tests N time (e.g `go test -count N`); I prefer to not have these files around.

SimonKagstrom:
Not that I can remember, although it was a long time since I initially set this up, so feel free to relocate them as you wish!

PR #435, opened 2024-04-05
Additional test fixes

perillo:
- [x] tests: fix incorrect code in `test_compiled`.py
    
    The `SignalBase` method of the `SignalBase` class is never called; probably it should have be the `__init__` method.  Remove it, since `self.m_self` is always set on subclasses.
    
    In `collect_and_report_only`, use `Exception`, instead of `BaseException`, since the former does not catch `GeneratorExit`, `KeyboardInterrupt` and `SystemExit`.
    
    In `system_mode_can_record_and_report_binary`, the test fails as expected, but the cause of the error is:
        NameError: name 'time' is not defined
    
    Import the `time` module.
    
    Errors reported by `ruff check`.

  - [x] tests: make naming convention consistent in `test_compiled.py`
    
    Rename `Pie` to `pie`.
    
    I forgot to apply this change in d2403a7 (tests: use consistent naming convention in test cases).

  - [x] tests: fix typo in `test_python.py`
    
    Rename `python_coverage.runTestTest` to `python_coverage.runTest`.

  - [x] tests: fix `test_compiled.address_sanitizer_coverage` test
    
    Restore building of the `sanitizer-coverage` executable in cmake, and just skip the test if it was not built with clang.
    
    The original code made it hard to understand the purpose of the test.

  - [x] tests: remove obsolete code from `tests/CMakeLists`
    
    cmake 3.12 supports the `POSITION_INDEPENDENT_CODE` property, so there is no need for the version check.

## NOTES

I'm not fully sure about commit "tests: fix test_compiled.address_sanitizer_coverage test".


SimonKagstrom:
Thanks, looks good!

The sanitizer coverage collection is also not completely working, and is probably not very interesting anyway (since it still requires special compile options, which is sort of the point of kcov to start with).


And kudos for the very good PR descriptions by the way!

perillo:
> Thanks, looks good!
> 
> The sanitizer coverage collection is also not completely working, and is probably not very interesting anyway (since it still requires special compile options, which is sort of the point of kcov to start with).
> 

Yes, I was aware that the sanitizer is not working correctly.  The problem is that the old test used `assert False` to force the test to fail.  Now the code is a bit more easy to understand.

> And kudos for the very good PR descriptions by the way!

Thanks!
Also, thank you for taking your time to review all the changes that I proposed.

PR #440, opened 2024-07-06
CI upgrades and docker setup

williamdes:
This is marked as draft but you can merge my first CI commits

The action https://github.com/dawidd6/action-get-tag is deprecated, I replaced it with logic I use at work.


SimonKagstrom:
Looks good to me!

Perhaps the docker image would be better to switch to alpine while we're at it? At least my Raspberry pi images built on alpine tends to be smaller than the debian-based ones.

williamdes:
It looks like s390x does not build
```
  > [linux/s390x builder 4/4] RUN mkdir /src/build &&     cd /src/build &&     cmake -G 'Ninja' .. &&     cmake --build . &&     cmake --build . --target install:
20.40       |   ^~~~~
20.40 /src/src/solib-parser/lib.c:105:25: error: expected string literal before ')' token
20.40   105 |                         );
```
https://github.com/williamdes/kcov/actions/runs/9819307965/job/27112899974#step:6:10565

SimonKagstrom:
Didn't know it tried to build for s390!

Anyway, kcov only supports x86, arm, aarch64/arm64, riscv and loongarch, so no need to build for s390 etc.

williamdes:
> Didn't know it tried to build for s390!

It's not yet on my PR


> Anyway, kcov only supports x86, arm, aarch64/arm64, riscv and loongarch, so no need to build for s390 etc.

Well, I have a s390x VM at IBM at Linux One. 
It would be nice to have it supported. 

But yeah, let's drop some old archs. Even if this hurts me a bit but they are probably not used and wasting power to build them is no good. 

It's

SimonKagstrom:
They are still built and tested (well, not MIPS), just not the docker image.

As for s390, it needs a few lines for the breakpoint encoding. Probably not difficult to add, but I don't really have a good way of testing it myself (and no personal use for it either).

williamdes:
> They are still built and tested (well, not MIPS), just not the docker image.
> 
> As for s390, it needs a few lines for the breakpoint encoding. Probably not difficult to add, but I don't really have a good way of testing it myself (and no personal use for it either).

Do you want me to add your ssh key to my Debian 12 VM?
Email me at williamdes / wdes.fr

SimonKagstrom:
> > As for s390, it needs a few lines for the breakpoint encoding. Probably not difficult to add, but I don't really have a good way of testing it myself (and no personal use for it either).
> 
> Do you want me to add your ssh key to my Debian 12 VM? Email me at williamdes / wdes.fr

Thanks for the offer, but since I won't really use it myself it would be better for someone else to implement and add it. Also, I guess the s390 architecture might be special in other ways as well, so maybe it's not as the others, which are basically

```
[...]
#elif defined(__powerpc__)
	val = 0x7fe00008; /* tw */
#elif defined(__arm__)
	val = 0xfedeffe7; // Undefined insn
[...]
```
(and slightly more involved for x86 etc).

williamdes:
Same error in mips64le: https://github.com/SimonKagstrom/kcov/pull/440#issuecomment-2211818081
Logs: https://github.com/williamdes/kcov/actions/runs/9821201170/job/27116829723#step:6:9314

I am testing all archs supported by the Debian image and commenting out the unsupported ones. One by one.
So we have a list at today's date.

SimonKagstrom:
Great, thanks for all the effort!

williamdes:
Build success of: `linux/386,linux/amd64,linux/arm/v5,linux/arm/v7,linux/arm64/v8,linux/ppc64le`
See: https://github.com/williamdes/kcov/actions/runs/9821655706/job/27117558504#step:6:9856

williamdes:
> Perhaps the docker image would be better to switch to alpine while we're at it? At least my Raspberry pi images built on alpine tends to be smaller than the debian-based ones.

Could we merge and test all this before switching to Alpine ?

williamdes:
I added an "environment" name, it has some benefits: 
- Limit the access to secrets to only some workflows
- You can add security filters on branches and tags in the UI
- It makes a nice UI to track deploys (https://github.com/williamdes/kcov/deployments)

You are not obliged to move the existing secrets into it, no change is required at all.
But when this PR is merged, feel free to put the secrets into an env and maybe use a token and not your password in the password env.

image

SimonKagstrom:
> > Perhaps the docker image would be better to switch to alpine while we're at it? At least my Raspberry pi images built on alpine tends to be smaller than the debian-based ones.
> 
> Could we merge and test all this before switching to Alpine ?

Absolutely!

I'll take a look at this tomorrow though, but thanks a lot again for the effort!

Dockerfile
```diff
@@ -23,19 +23,30 @@ RUN mkdir /src/build && \
...
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
-
-COPY --from=builder /usr/local/bin/kcov* /usr/local/bin/
-COPY --from=builder /usr/local/share/doc/kcov /usr/local/share/doc/kcov
+    ; \
+    apt-get clean; \
+    rm -rf /var/lib/apt/lists/*; \
+    # Write a test script
+    echo '#!/usr/bin/env bash\nif [[ true ]]; then\necho "Hello, kcov!"\nfi' > /tmp/test-executable.sh; \
```
SimonKagstrom, line 41: This script is also available as `.github/workflows/test-executable.sh`, so might be possible to simply copy from there.

However, I think a better test (although more fickle) would be to test against a compiled binary, since the shell script code isn't architecture-specific. That would require the docker builder to run with `--security-opt seccomp=unconfined` though, since ptrace normally isn't available via docker.

williamdes: > This script is also available as .github/workflows/test-executable.sh, so might be possible to simply copy from there.

Yes but if I do that it will be part of the final image, not very clean..

> However, I think a better test (although more fickle) would be to test against a compiled binary, since the shell script code isn't architecture-specific. That would require the docker builder to run with --security-opt seccomp=unconfined though, since ptrace normally isn't available via docker.

Do you have a simple binary in mind ?

SimonKagstrom: A simple

```
int main(int argc, const char *argv[])
{
   printf("Hello world!\n");
   if (argc > 1000)
   {
      printf("Unacceptable!\n");
   } 
   return 0;
}
```

will do just fine. It's basically how the automated tests are done, where the cobertura file is parsed to verify that e.g., the "Hello world" line is executed, but that the "Unacceptable!" line isn't. There's a helper script, `tests/tools/parse_cobertura` which does the lookup, given a file and line pair.

It doesn't work for the non-x86 github action runners since these run via qemu-user, where ptrace isn't possible to use. They instead just check that a basic shell script works, like in your Dockerfile.

For docker images, one instead needs the `--security-opt` mentioned above.

williamdes:
It looks like I can not build the current code on Alpine
```
#5 1.900 /src/src/engines/ptrace_linux.cc: In function 'long int getRegs(pid_t, void*, void*, size_t)':
#5 1.900 /src/src/engines/ptrace_linux.cc:332:24: error: '__ptrace_request' was not declared in this scope
#5 1.900   332 |         return ptrace((__ptrace_request ) PTRACE_GETREGS, pid, NULL, regs);
#5 1.900       |                        ^~~~~~~~~~~~~~~~
#5 1.900 /src/src/engines/ptrace_linux.cc: In function 'long unsigned int ptrace_sys::peekWord(pid_t, long unsigned int)':
#5 1.900 /src/src/engines/ptrace_linux.cc:364:24: error: '__ptrace_request' was not declared in this scope
#5 1.900   364 |         return ptrace((__ptrace_request ) PTRACE_PEEKTEXT, pid, aligned_addr, 0);
#5 1.900       |                        ^~~~~~~~~~~~~~~~
#5 1.900 /src/src/engines/ptrace_linux.cc: In function 'void ptrace_sys::pokeWord(pid_t, long unsigned int, long unsigned int)':
#5 1.900 /src/src/engines/ptrace_linux.cc:369:17: error: '__ptrace_request' was not declared in this scope
#5 1.900   369 |         ptrace((__ptrace_request ) PTRACE_POKETEXT, pid, aligned_addr, value);
#5 1.900       |                 ^~~~~~~~~~~~~~~~
#5 1.900 /src/src/engines/ptrace_linux.cc: In function 'long int setRegs(pid_t, void*, void*, size_t)':
#5 1.900 /src/src/engines/ptrace_linux.cc:380:24: error: '__ptrace_request' was not declared in this scope
#5 1.900   380 |         return ptrace((__ptrace_request ) PTRACE_SETREGS, pid, NULL, regs);
#5 1.900       |                        ^~~~~~~~~~~~~~~~
```

It looks like this is something to fix in the code ?

williamdes:
In fact this is needed
`CXXFLAGS="-D__ptrace_request=int" cmake -G 'Ninja' ..`
Not sure maybe the code could be adjusted to avoid this ?

SimonKagstrom:
Yes, I think the code should be adjusted instead. Not actually sure the cast is needed at all. The FreeBSD port doesn't use it, for example.

So a first test could be to simply remove all of `(__ptrace_request)`.

PR #441, opened 2024-07-07
Move to Alpine 3.20

williamdes:
Not sure why but my simple test is broken
```
#10 2.970   "files": [
#10 2.970 
#10 2.970   ],
#10 2.970   "percent_covered": "0.00",
#10 2.970   "covered_lines": 0,
#10 2.970   "total_lines": 0,
#10 2.970   "percent_low": 25,
#10 2.970   "percent_high": 75,
#10 2.970   "command": "test-executable.sh",
#10 2.970   "date": "2024-07-07 17:57:41"
```

No percent and no file reported

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ ADD . /src/
 RUN mkdir /src/build && \
     cd /src/build && \
     export PATH="$PATH:/usr/lib/ninja-build/bin/" && \
-    cmake -G 'Ninja' .. && \
+    CXXFLAGS="-D__ptrace_request=int" cmake -G 'Ninja' .. && \
     cmake --build . && \
     cmake --build . --target install
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,42 +1,48 @@
-FROM debian:bookworm-slim AS builder
+FROM alpine:3.20 AS builder
 
-RUN apt-get update && \
-    apt-get install -y \
+RUN apk add --update --no-cache \
         binutils-dev \
-        build-essential \
+        # Debian: build-essential
+        build-base \
         cmake \
         git \
-        libcurl4-openssl-dev \
-        libdw-dev \
-        libiberty-dev \
-        libssl-dev \
+        # Debian: libcurl4-openssl-dev
+        curl-dev \
+        curl-static \
+        # Debian: libdw-dev
+        libdw \
+        # Debian, alpine bundled in binutils-dev
+        # libiberty-dev \
+        # Debian: libssl-dev
+        openssl-dev \
         ninja-build \
         python3 \
-        zlib1g-dev \
-        ;
+        # Debian: zlib1g-dev
+        zlib-dev \
+        # Debian: libelf-dev
+        elfutils-dev \
+    ;
 
 ADD . /src/
 
 RUN mkdir /src/build && \
     cd /src/build && \
+    export PATH="$PATH:/usr/lib/ninja-build/bin/" && \
     cmake -G 'Ninja' .. && \
     cmake --build . && \
     cmake --build . --target install
 
-FROM debian:bookworm-slim
+FROM alpine:3.20
 
 COPY --from=builder /usr/local/bin/kcov* /usr/local/bin/
 COPY --from=builder /usr/local/share/doc/kcov /usr/local/share/doc/kcov
 
 RUN set -eux; \
-    apt-get update; \
-    apt-get install -y --no-install-recommends --no-install-suggests \
-        libcurl4 \
-        libdw1 \
-        zlib1g \
+    apk add --update --no-cache \
+        libcurl \
+        libdw \
+        zlib \
     ; \
-    apt-get clean; \
-    rm -rf /var/lib/apt/lists/*; \
     # Write a test script
     echo '#!/usr/bin/env bash\nif [[ true ]]; then\necho "Hello, kcov!"\nfi' > /tmp/test-executable.sh; \
     # Test kcov
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ ADD . /src/
 RUN mkdir /src/build && \
     cd /src/build && \
     export PATH="$PATH:/usr/lib/ninja-build/bin/" && \
-    CXXFLAGS="-D__ptrace_request=int" cmake -G 'Ninja' .. && \
+    cmake -G 'Ninja' .. && \
     cmake --build . && \
     cmake --build . --target install
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,8 @@ RUN apk add --update --no-cache \
         zlib-dev \
         # Debian: libelf-dev
         elfutils-dev \
+        # Debian: libstdc++-12-dev
+        libstdc++-dev \
     ;
 
 ADD . /src/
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,42 +1,53 @@
...
+        libdw \
+        zlib \
+        libgcc \
+        libstdc++ \
     ; \
-    apt-get clean; \
-    rm -rf /var/lib/apt/lists/*; \
     # Write a test script
     echo '#!/usr/bin/env bash\nif [[ true ]]; then\necho "Hello, kcov!"\nfi' > /tmp/test-executable.sh; \
     # Test kcov
```
SimonKagstrom, line 53: You can remove the --include-pattern here. Since it's only one file, it's not really needed.

Since it's a bash script, you could consider adding `--bash-dont-parse-binary-dir` though, so that it doesn't try to find other files to cover in /tmp/

williamdes: None of the two options do work, what could be wrong on Alpine ?

SimonKagstrom: Since it's bash coverage, is bash actually installed in the image?

Most likely explanation is otherwise that there is some incompatibility in the version used in the docker image (i.e., most likely a newer bash version in Alpine).

SimonKagstrom: An uncommon option which disables the xtrace file descriptor is:

```--configure=bash-force-stderr-input=1```

not sure, but maybe could be the cause of the alpine issues. There's also `--bash-method=DEBUG`, but that sounds less likely.

Comment (williamdes) [reply Dockerfile:53]: > Since it's bash coverage, is bash actually installed in the image?

Yes, I added a line or the script exec failed.

Okay, pushing `--configure=bash-force-stderr-input=1` to try

Comment (SimonKagstrom) [reply Dockerfile:53]: Sorry, was looking in the source code.

`--debug-force-bash-stderr` should be the option. If that doesn't work, then perhaps we should let Alpine be?

However, would be nice to build docker images for riscv as well, which I also saw that you added  here (even if it's for bookworm)

Comment (williamdes) [reply Dockerfile:53]:
```json
 {
  "files": [
     {"file": "/tmp/coverage/bash-helper-debug-trap.sh", "percent_covered": "0.00", "covered_lines": "0", "total_lines": "3"},
     {"file": "/tmp/coverage/bash-helper.sh", "percent_covered": "0.00", "covered_lines": "0", "total_lines": "2"}
   ],
   "percent_covered": "0.00",
   "covered_lines": 0,
   "total_lines": 5,
   "percent_low": 25,
   "percent_high": 75,
  "command": "test-executable.sh",
  "date": "2024-07-08 10:50:40"
 }
```

SimonKagstrom: I tried pasting the commands locally and tested.

I wonder if  something is wrong there: At least I get no line endings with the command, which gives the same results as you see. If I manually edit the file, it works correctly:

```
Simons-iMac:emilpro ska$ cat /tmp/kcov/test-executable.sh/coverage.json
{
  "files": [

  ],
  "percent_covered": "0.00",
  "covered_lines": 0,
  "total_lines": 0,
  "percent_low": 25,
  "percent_high": 75,
  "command": "test-executable.sh",
  "date": "2024-07-08 13:11:52"
}
Simons-iMac:emilpro ska$ cat /tmp/test-executable.sh
#!/usr/bin/env bash\nif [[ true ]]; then\necho "Hello, kcov!"\nfi

[... edit the executable to get correct line endings ]

Simons-iMac:emilpro ska$ kcov --bash-dont-parse-binary-dir /tmp/kcov /tmp/test-executable.sh
Hello, kcov!
Simons-iMac:emilpro ska$ cat /tmp/test-executable.sh
#!/usr/bin/env bash
if [[ true ]]; then
echo "Hello, kcov!"
fi
Simons-iMac:emilpro ska$ cat /tmp/test-executable.sh
#!/usr/bin/env bash
if [[ true ]]; then
echo "Hello, kcov!"
fi
Simons-iMac:emilpro ska$ cat /tmp/kcov/test-executable.sh/coverage.json
{
  "files": [
    {"file": "/private/tmp/test-executable.sh", "percent_covered": "100.00", "covered_lines": "2", "total_lines": "2"}
  ],
  "percent_covered": "100.00",
  "covered_lines": 2,
  "total_lines": 2,
  "percent_low": 25,
  "percent_high": 75,
  "command": "test-executable.sh",
  "date": "2024-07-08 13:12:20"
}
```

Dockerfile
```diff
@@ -1,46 +1,57 @@
...
+        libcurl \
+        libdw \
+        zlib \
+        libgcc \
+        libstdc++ \
     ; \
-    apt-get clean; \
-    rm -rf /var/lib/apt/lists/*; \
     # Write a test script
     echo '#!/usr/bin/env bash\nif [[ true ]]; then\necho "Hello, kcov!"\nfi' > /tmp/test-executable.sh; \
```
SimonKagstrom, line 52: The ' probably messes this up. I think the best might be to use multi-line-syntax, e.g., something like

```
cat <<EOF > /tmp/test-executable.sh
#!/usr/bin/env bash
if [[ true ]]; then
    echo "Hello, kcov!"
fi
EOF
```

Dockerfile
```diff
@@ -1,46 +1,57 @@
...
+        libstdc++-dev \
+    ;
 
 ADD . /src/
 
 RUN mkdir /src/build && \
     cd /src/build && \
-    cmake -G 'Ninja' .. && \
+    export PATH="$PATH:/usr/lib/ninja-build/bin/" && \
+    CXXFLAGS="-D__ptrace_request=int" cmake -G 'Ninja' .. && \
```
SimonKagstrom, line 33: I made a PR to remove this, and tagged you in the comments.

After that the CXXFLAGS part can/should be removed.

Dockerfile
williamdes: ah yes !
Good catch, I need to add -e to echo.

the EOF stuff does not work fine in the Dockerfile syntax..

Dockerfile
williamdes: Okay, I will rebase when #443 is merged

Dockerfile
williamdes: `--include-pattern` is needed or coverage drops

SimonKagstrom: Yes, that's fine. I think it's due to it finding other files in /tmp/ though, so `--bash-dont-parse-binary-dir` should also give 100%.

But go with --include-pattern, it's clear enough!

williamdes:
Docker is building at: https://github.com/williamdes/kcov/actions/runs/9842731065/job/27172308187
Should be ready to merge

williamdes:
By the way libcurl seems to have some deprecation notices: https://github.com/williamdes/kcov/actions/runs/9842731065/job/27172308187#step:6:1266

williamdes:
All good, build passed: https://github.com/williamdes/kcov/actions/runs/9842731065/job/27172308187#step:6:2544

williamdes:
About the example binary, if you can add it after the merge to the existing CI scripts I can copy the login into the Dockerfile

SimonKagstrom:
Thanks for the PR, great work!

Don't worry about the binary. I suspect it won't be possible to run anyway in the docker image, at least on non-x86 images, so I think the current variant is good enough. It's tested in the regular CI runs anyhow.

williamdes:
Thank you for the very active and nice collaboration !
I updated https://github.com/sudo-bot/action-kcov

PR #442, opened 2024-07-07
ci: Add riscv

SimonKagstrom:


SimonKagstrom:
@williamdes your work for dockerhub made me remember this branch I had laying around. It had a lot of other changes to run tests as well, but those doesn't work anyway. So now it's just adding riscv to the github ci.

williamdes (approved):
Typo on the commit message
Also, it's not possible to enable riscv64 for bookworm images because the Docker hub image is not build for riscv64 itself. But alpine does have it.

SimonKagstrom:
Sorry, merged before I had read the comment :-/

I guess it has to be known as a 62 bit CPU in kcov!


Anyway, this is not for the docker builds, so only to verify that it's possible to build in general.

williamdes:
> Sorry, merged before I had read the comment :-/
> 
> I guess it has to be known as a 62 bit CPU in kcov!
> 
> Anyway, this is not for the docker builds, so only to verify that it's possible to build in general.

Yup, that's all good
Nice we have a check for it

williamdes:
Can you add a musl VM?

SimonKagstrom:
> Can you add a musl VM? (Alpine)

Not quite sure how to do that? You mean for the github actions as a build job? With some googling, this turned up, https://github.com/jirutka/setup-alpine, maybe something like that?

williamdes:
> > Can you add a musl VM? (Alpine)
> 
> Not quite sure how to do that? You mean for the github actions as a build job? With some googling, this turned up, https://github.com/jirutka/setup-alpine, maybe something like that?

It definitely looks like the way to do this !

Another way is to set the image in a pure Docker way
```yaml
    job-name:
        container:
            image: alpine:3.20
        runs-on: ubuntu-latest
        steps:
            - name: Checkout repository
```

PR #443, opened 2024-07-08
Linux correct ptrace calls

SimonKagstrom:


williamdes:
> Unrelated to that, I added DOCKER_USERNAME+DOCKER_PASSWORD to the environment secrets now. Is there anything else to do, or will that work by magic after that?

Nothing more, you should not see the secrets anymore in the normal secret list now.
Did you use a token and not a password ?

I hesitated to use vars.DOCKER_USERNAME but it would have broken the setup.
Feel free to move the username to a variable as it is not sensitive and can help with debug sometimes.

williamdes (approved):

SimonKagstrom:
Didn't move to a token yet, need to readup a bit more first!

williamdes:
> Didn't move to a token yet, need to readup a bit more first!

Sure
https://hub.docker.com/settings/security

Issue #444, opened 2024-07-08
Debian packaging


williamdes:
I noticed there is a debian/ folder in this repo (6 years old: https://github.com/SimonKagstrom/kcov/commits/master/debian)
But the package https://tracker.debian.org/pkg/kcov
Has a git repo in Debian Salsa: https://salsa.debian.org/debian/kcov
Is there a reason for this folder here ?

I will have to contact [Alessandro Ghedini](https://qa.debian.org/developer.php?email=ghedo%40debian.org) and try to package the latest version

SimonKagstrom:
It was a contribution due to the (back then) outdated debian package. I don't think it's relevant anymore, so good to remove!

It's a bit unfortunate that the debian packages are a bit behind, and thanks for taking it up with the maintainer!

williamdes:
I got access to the Salsa repo, will be updating kcov in Debian.

Can you see if the patch in https://salsa.debian.org/debian/kcov/-/merge_requests/1/diffs can be merged upstream (here) ?

SimonKagstrom:
Thanks!

I manually merged the patch from debian above.

Anyway, I suppose a release should be done before the update in Debian?

williamdes:
> Thanks!
> 
> I manually merged the patch from debian above.
> 
> Anyway, I suppose a release should be done before the update in Debian?

Yes, but I will probably trick the system to test that everything work before the release of v43

williamdes:
Thank you so much for https://github.com/SimonKagstrom/kcov/commit/a7ab2971d377e6df84af52b748d46fa40d18f6c8

williamdes:
Could you GPG sign your tags in the future ?
And git commits too ?

williamdes:
I think I need some help

On Debian I am trying to build tests. `cmake ../tests/unit-tests`

here is the error:
```
CMake Error at /usr/share/cmake-3.25/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find LibCRPCUT (missing: LIBCRPCUT_LIBRARIES
  LIBCRPCUT_INCLUDE_DIRS)
Call Stack (most recent call first):
  /usr/share/cmake-3.25/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  /mnt/Dev/@debian/@others/kcov/cmake/FindLibCRPCUT.cmake:54 (find_package_handle_standard_args)
[...]

Issue #437, opened 2024-05-06
KCOV report-only not working as expected Mac OS


9-para:
`kcov --collect-only ./coverage ./test.sh` working as expected tried with `--debug` looks good collecting & generating data. 

While use the `--report-only` it's appending hash at the end of filename which does not exist in directory. 

`kcov --report-only ./coverage ./test.sh` tried deleting coverage & regenerating along with `--clean` however each time it's creating same **Hash**.

SimonKagstrom:
I think this might be an issue with bash coverage (and I guess Python will show the same problem). I've really only used collect/report with compiled code, so I don't think it's really an issue specific to MacOS.

9-para:
You are right it's not issue specific to MacOS it's with bash coverage.

SimonKagstrom:
Having investigated this further, there seems to be two issues: One is that the hash given to the filename in some instances uses the file content, and in some the file name, which explains the problem when running report-only.

However, I tried removing that part, and it still doesn't work, so there's more to it than that. Probably there are more instances where there's a mixup between filenames and contents, but I don't really know.

williamdes:
```
 kcov v41-161-gce4b
Ran 1 test.
kcov: error: report-only selected, but the target directory /tmp/kcov//test.sh.242af937 does not exist
```
It looks like I ran into this one
Logs: https://github.com/sudo-bot/action-shunit2/actions/runs/9846037080/job/27182922365#step:4:6

Previous versions worked fine:
```
 kcov 38
Ran 1 test.
-rw-r--r--    1 root     root        3.7K Jul  8 19:47 /tmp/kcov/index.html
```
Logs: https://github.com/sudo-bot/action-shunit2/actions/runs/9845959366/job/27182662377#step:4:8

SimonKagstrom:
I believe this was caused by the changes made for #343 , but I haven't really figured out what.

SimonKagstrom:
PR-448 should get this back to how v38 worked. However, even that is not really working for bash scripts, and for me, I don't get coverage information with v38 either. I suspect it has really never worked with shell scripts.

The original motivation for this was to be able to use `--collect-only` on a "target" machine, which has the binaries, and `--report-only` on a machine where the source code resides. But that usecase is only relevant for compiled code.

SimonKagstrom:
I'll leave this open, since it only fixes the `kcov: error: report-only selected, but ...` error, not the actual collect + report sequence for bash scripts.

[...]

PR #445, opened 2024-07-08
Documentation improvements and drop debian/

williamdes:
Ref #444


src/CMakeLists.txt
```diff
@@ -57,6 +57,7 @@ if (EXISTS "${CMAKE_SOURCE_DIR}/.git")
             "--git-dir=${CMAKE_SOURCE_DIR}/.git"
             describe
             --abbrev=4
+            --tags
```
williamdes, line 60: The last tag was not annotated, this fixes it
```diff
- v41-167-gfa5b
+ v42-136-gfa5b
```

williamdes:
If you can add the following uncommon options to the kcov.1 file, that would help me not having to write in this uncommon format ?

I checked help output and the file
Missing:
```
 --gcov                  use gcov parser instead of DWARF debugging info
 --clang                 use Clang Sanitizer-coverage parser
 --dump-summary          dump a summary of coverage on stdout

 --output-interval=ms    Interval to produce output in milliseconds (0 to
                         only output when kcov terminates, default 5000)

 --debug=X               set kcov debugging level (max 31, default 0)

 --configure=key=value,... Manually set configuration values. Possible values:
                           bash-use-basic-parser=1    Enable simple bash parser
                           command-name=STR           Name of executed command
                           css-file=FILE              Filename of bcov.css file
                           high-limit=NUM             Percentage for high coverage
                           low-limit=NUM              Percentage for low coverage
                           merged-name=STR            Name of [merged] tag in HTML
                           coveralls-service-name=STR Service name for coveralls

 --patchelf=cmd          patchelf binary to use (for system mode),
                         default: patchelf

 --bash-tracefd-cloexec  Force children to not be traced by configuring the trace
                         fd as non-cloneable with LD_PRELOAD. Buggy on some systems

```

src/CMakeLists.txt
SimonKagstrom: Ah, OK!

I guess it's soon time for a new release again.

SimonKagstrom:
Now merged the documentation changes as #446, thanks for noting it!

williamdes:
> Now merged the documentation changes as #446, thanks for noting it!

Thank you !

Only #437 is left before trying to update this into Debian

PR #448, opened 2024-07-09
Issue #437: Fixes for --collect-only/--report-only

SimonKagstrom:
For bash, this corrects the name mismatch between --collect-only/--report-only. It doesn't work even then, and I don't think it ever has.

For Mach-O (MacOS binaries) it corrects --report-only.

### DIFF
--- a/doc/vscode.md
+++ b/doc/vscode.md
@@ -8,7 +8,7 @@ extension when the coverage changes.
 
 To produce this output, just run
 
-```
+```sh
 kcov --cobertura-only [other options] /path/to/your/folder/.vscode /path/to/the/binary
 ```
 
--- a/src/configuration.cc
+++ b/src/configuration.cc
@@ -493,11 +493,26 @@ public:
 			setKey("binary-path", tmp.first);
 			binaryName = tmp.second;
 
+			setKey("binary-name", binaryName);
+
 			if (!keyAsInt("cobertura-only"))
 			{
-				setKey("target-directory",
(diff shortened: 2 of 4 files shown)


SimonKagstrom:
#437 should techically be fixed by this, although it still doesn't work as compiled code does.

williamdes:
> #437 should techically be fixed by this, although it still doesn't work as compiled code does.

What is the difference in behaviour with this fix?

williamdes:
Maybe we should add to the testsuite and docs my shunit2 example
as it is a great application of kcov

SimonKagstrom:
> > #437 should techically be fixed by this, although it still doesn't work as compiled code does.
> 
> What is the difference in behaviour with this fix?

The regression was that there were two different names used for the "binary" directory: One being a hash of the filename, and the other a hash of the file contents. The fix uses content for both of them, which fixes the `error: report-only selected, but the target ...` error.

However, the output is still empty and I don't think bash collect+report has ever really worked.


The testsuite needs some overhaul (well, most of the code does), but the main issue with bash has always been that it doesn't use a proper parser, so it's easy to get lost while parsing. Until that's fixed, I think we can hold further bash testsuite improvements. Unfortunately, I'm quite frankly unlikely to do any work on that, since it's a quite complex task unless there's an easy-to-use C or C++ library that parses it. Also, I really only use kcov for compiled code myself :-)

SimonKagstrom:
> Maybe we should add to the testsuite and docs my shunit2 example as it is a great application of kcov

Replied in the other comment, but clarification here :-). Yes, that would be a good addition! However, I think it will be more important once the parser is better.

Issue #457, opened 2024-07-21
Bash: heredoc and arrays are not correctly covered


beatussum:
# Description

In bash, if a heredoc has space between the tag and `<<[-]`, the lines of the string are not covered.

Same thing with arrays if they are not defined oneline.

It occurs also with functions in subshells like:

```bash
foo() (
    echo "bar"
)
```

# Example

![Example](https://github.com/user-attachments/assets/9517238b-bf6a-44a4-bb84-5693550d0d5c)

SimonKagstrom:
Thanks for the clear test!

This, and other bash parsing issues, should really be handled by #145 , but I don't really have any timeframe for that. I mostly (almost only) use kcov for compiled code myself, so the bash support is sort of neglected by me.

beatussum:
Currently, the following has been _fixed_ (it's not perfect and other issues risk to occur):
- [x] code coverage of _heredoc_ having spaces between `<<[-]` and the tag (see e208a4ed2180c70dc94b84faf438ffedbc277765 provided by #459).
- [x] code coverage of subshell functions (see 32c6189c8fd88a164d381fdb77ec6350153923ea provided by #459).
- [ ] code coverage of multiline arrays. As it's explained in [this comment](https://github.com/SimonKagstrom/kcov/pull/459#issuecomment-2261347586), parsing of these arrays seems quiet difficult without a true parser (see #145).

SimonKagstrom:
Thanks a lot for these! I'm closing this issue.

Another note is that there is also a basic parser for bash scripts, which can be used if the default parser goes astray, so I'm not to worried about additional issues.

PR #459, opened 2024-07-31
Fix some issues in the Bash engine

beatussum:
This pull request refers to [this issue](https://github.com/SimonKagstrom/kcov/issues/457) I reported some days ago.

It's not perfect but it should do the job in most of the cases.

I tried to respect your coding style. I hope I haven't made mistake and I'm waiting for review.

All tests passed.

### DIFF
--- a/src/engines/bash-engine.cc
+++ b/src/engines/bash-engine.cc
@@ -756,8 +756,17 @@ private:
 				if (!arithmeticActive && s.find("let ") != 0 && s.find("$((") == std::string::npos
 						&& s.find("))") == std::string::npos && heredocStart != std::string::npos)
 				{
-					// Skip << and remove spaces before and after "EOF"
-					heredocMarker = trim_string(s.substr(heredocStart + 2, s.size()));
+					// Skip <<
+					heredocMarker = s.substr(heredocStart + 2, s.size());
+
+					if (heredocMarker[0] == '-')
+					{
+						// '-' marks tab-suppression in heredoc
+						heredocMarker = heredocMarker.substr(1);
+					}
+
+					// Remove spaces before and after "EOF"
+					heredocMarker = trim_string(heredocMarker);
 
 					// Make sure the heredoc marker is a word
 					for (unsigned int i = 0; i < heredocMarker.size(); i++)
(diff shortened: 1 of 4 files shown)


SimonKagstrom:
Thanks!

You had a great example in the issue, is it possible to add that to a small test case?


Bash parsing is fairly nasty. The python support is much more slim, and easier to follow, but bash requires quite a lot of special cases. I had hopes of writing a proper parser, but with these changes (and some other), it's probably good enough as it is!

And as for the coding style, I've started to change it in some files, so right now everything is in flux and your changes look good to me :-)

beatussum:
I've dropped the commit which was supposed to fix coverage of multiline arrays. Indeed, it seems to be quiet difficult as `trap <arg> DEBUG` does not have the same behavior with the different syntaxes used to initialize an array:

```bash
trap 'echo ${LINENO}' DEBUG

declare -a array=( # <- this line
    a
    b
    c
)

declare array=(
    a
    b
    c
) # <- this line

declare array=(
    a
    b
    c
) # <- this line
```

SimonKagstrom:
Merged, thanks a lot!

williamdes:
Hello here

I applied https://github.com/SimonKagstrom/kcov/commit/a6ff5ac44524d1bc9c298ca14c10edd41ded1cea on the v43 
And I get this test error

So I tried applying most of this pull-request, but still it fails on this test.
On my workstation it works fine, but on Debian testing it does not. Why, I have no idea.

Can you help me ?

```
======================================================================
FAIL: runTest (test_bash.bash_subshell.runTest)
Traceback (most recent call last):
  File "/builds/debian/kcov/debian/output/source_dir/tests/tools/test_bash.py", line 353, in runTest
    self.assertEqual(2, cobertura.hitsPerLine(dom, "subshell.sh", 4))
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: 2 != 0

```

SimonKagstrom:
@williamdes I wonder if that actually is an issue with the test. It might be  that a later version of Debian containing a newer bash might report the line differently. It's this piece of code:

```
#/bin/sh
eval "(echo Stuff in subshell
echo More stuff in subshell
      ) "
```

where the last line is expected to be marked as executed, which the older bashes did. Perhaps the newer versions instead report the first line?

PR #460, opened 2024-08-19
license: Move to SPDX expressions

dridi:
This clarifies which flavor of GPLv2 is being used, in particular for the kcov code itself since it's never mentioned whether it is "or later" or strictly this version.


SimonKagstrom:
Thanks, merged!

New concept for me as well!

Issue #458, opened 2024-07-30
macos: kcov enters infinite loop when run with bats executable with bash 3.2 (/bin/bash)


twils0:
When I run kcov (as detailed below), it enters an infinite loop. I can't seem to get kcov to honor my exclude/include flags. I've tried many different variations of {include|exclude}-{path|pattern}.

I installed executables for kcov and bats in my project. I want to get the coverage for script.sh, based on the bats test script_test.bats.

Here's an example of the commands I'm tried:

`kcov/bin/src/kcov --bash-dont-parse-binary-dir --exclude-pattern=bats/ --include-pattern=script.sh coverage/ ./bats/bin/bats script_tests.bats`

Here's my project structure:

```
project/
project/script.sh
project/script_test.bats
project/bats/bin/bat    <= bats executable
project/bats/bats-core
project/bats/bats-support
project/bats/bats-assert
project/kcov/*  <= cloned from Github
project/kcov/bin/src/kcov    <= kcov executable
```

Here's the last few lines of the infinite loop that kcov enters into. I can't get kcov to ignore these files related to the bats executable.

I've also downloaded bats through Homebrew. When I run kcov against the bats executable from Homebrew, this infinite loop again occurs with paths pointing at the bats Homebrew directory, instead of the bats directory in my project (as shown below).

```
# kcov@project/bats/lib/bats-core/common.bash@5@printf '# %s\n' 'kcov@project/bats/lib/bats-core/common.bash@5@printf '\''# %s\n'\'' '\''kkcov@project/bats/lib/bats-core/tracing.bash@281@for path in "${BATS_DEBUG_EXCLUDE_PATHS[@]}"'\'''
# kcov@project/bats/lib/bats-core/common.bash@15@printf '%s\n' '# kcov@project/bats/lib/bats-core/common.bash@15@printf '\''%s\n'\'' '\''# kkcov@project/bats/lib/bats-core/tracing.bash@281@for path in "${BATS_DEBUG_EXCLUDE_PATHS[@]}"'\'''
# kcov@project/bats/lib/bats-core/common.bash@14@read -r line
# kcov@project/bats/lib/bats-core/common.bash@15@printf '%s\n' '# kcov@project/bats/lib/bats-core/common.bash@4@IFS='
# kcov@project/bats/lib/bats-core/common.bash@14@read -r line
# kcov@project/bats/lib/bats-core/common.bash@4@IFS=
[...]

PR #466, opened 2024-09-23
Use $PATH to find Bash executable

jack-rann:
Fixes issue #458.  Use `execvp()` to find and execute the Bash interpreter, rather than defaulting to `/bin/bash`, which on many systems can be a old Bash implementation not supporting BASH_XTRACEFD.


jack-rann:
Note:  I will be out of town for the remainder of the week, so any response required of me will be delayed.

SimonKagstrom:
Looks good, thanks a lot for the fix!

PR #467, opened 2024-09-29
Handle Python execution via execvp()

jack-rann:
Modify `python-engine.cc` to handle sub-command execution in the same manner as `bash-engine.cc` (from PR #466).


SimonKagstrom:
Look good!

The ci failure seems to be because python2 is no longer available in Ubuntu 24.04 (?). Those tests might be moot now anyway, since python 3 has been the default for a long time.

PR #477, opened 2025-05-27
Add cmake to Ubuntu build packages

KlfJoat:
`cmake` is not available by default on a minimal installation of Ubuntu. Add it to the list of packages to install to build kcov.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -27,7 +27,7 @@ Ubuntu / Debian
 Install
 
 ```sh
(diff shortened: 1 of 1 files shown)


KlfJoat:
Also add build-essential to clear the following error when running `cmake`.

```text
CMake Error at CMakeLists.txt:2 (project):
  No CMAKE_CXX_COMPILER could be found.
```

SimonKagstrom:
Great, thanks

williamdes:
Noted, I think it needs a full review and testing

Issue #409, opened 2023-10-14
libcurl 7.56.0 deprecations


particleflux:
The usage of curl in the coveralls-writer gives a bunch of deprecation warnings when compiled against recent libcurl.

It would appear some of the `CURLFORM` things have been replaced with `curl_mime_*` functionality.

```
/var/tmp/portage/dev-util/kcov-42/work/kcov-42/src/writers/coveralls-writer.cc: In member function ‘bool CurlConnectionHandler::talk(const std::string&)’:
/var/tmp/portage/dev-util/kcov-42/work/kcov-42/src/writers/coveralls-writer.cc:65:33: warning: ‘CURLFORM_COPYNAME’ is deprecated: since 7.56.0. Use curl_mime_name() [-Wdeprecated-declarations]
   65 |                                 CURLFORM_COPYNAME, "json_file",
      |                                 ^~~~~~~~~~~~~~~~~
In file included from /var/tmp/portage/dev-util/kcov-42/work/kcov-42/src/writers/coveralls-writer.cc:16:
/usr/include/curl/curl.h:2510:3: note: declared here
 2510 |   CURLFORM_COPYNAME        CURL_DEPRECATED(7.56.0, "Use curl_mime_name()"),
      |   ^~~~~~~~~~~~~~~~~
/var/tmp/portage/dev-util/kcov-42/work/kcov-42/src/writers/coveralls-writer.cc:66:33: warning: ‘CURLFORM_FILE’ is deprecated: since 7.56.0. Use curl_mime_filedata() [-Wdeprecated-declarations]
   66 |                                 CURLFORM_FILE, fileName.c_str(),
      |                                 ^~~~~~~~~~~~~
/usr/include/curl/curl.h:2519:3: note: declared here
 2519 |   CURLFORM_FILE            CURL_DEPRECATED(7.56.0, "Use curl_mime_filedata()"),
      |   ^~~~~~~~~~~~~
/var/tmp/portage/dev-util/kcov-42/work/kcov-42/src/writers/coveralls-writer.cc:63:29: warning: ‘CURLFORMcode curl_formadd(curl_httppost**, curl_httppost**, ...)’ is deprecated: since 7.56.0. Use curl_mime_init() [-Wdeprecated-declarations]
   63 |                 curl_formadd(&formpost,
      |                 ~~~~~~~~~~~~^~~~~~~~~~~
   64 |                                 &lastptr,
      |                                 ~~~~~~~~~
[...]

PR #484, opened 2025-07-11
fix: use `curl_mime` in coveralls writer

supervacuus:
The `kcov` build currently generates deprecation warnings around the usage of `curl_formadd` (in the coveralls writer). This API has been deprecated in favor of the `curl_mime` API 8 years ago in curl `7.56.0`: https://github.com/curl/curl/pull/1839#issue-253654906.

This PR replaces the usage of `curl_formadd` with `curl_mime`.

Disclaimer: I haven't tested the integration with a Coveralls server. Maybe a user of that service could give this a test.

### DIFF
--- a/src/writers/coveralls-writer.cc
+++ b/src/writers/coveralls-writer.cc
@@ -86,21 +86,15 @@ public:
 	{
 		m_writtenData = "";
 
-		struct curl_httppost *formpost = NULL;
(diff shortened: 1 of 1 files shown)


SimonKagstrom:
Nice!

I belive this should fix #409 ?

supervacuus:
> I belive this should fix #409 ?

Yes, this fixes the issue described!

PR #495, opened 2026-04-01
fix: Preserve signal exit code when child process exits after crash

jpnurmi:
When a multi-process program crashes, kcov sometimes exits with zero instead of the signal number. When a child process exits normally _after_ the parent's signal exit has already been recorded, the child's `ev_exit` event overwrites the parent's exit code.

=> Keep track when `ev_signal_exit` sets the exit code so that subsequent `ev_exit` events from child processes cannot overwrite it.

See also:
- getsentry/sentry-native#1596
- getsentry/sentry-native#1419

Note: The `fork_crash` test relies on the child process outliving the parent's crash. The child sleeps for an arbitrary 5 seconds, which is way more than enough locally, but the test is not 100% guaranteed to reproduce the bug in loaded CI environments. Also, the `signals_crash` exit code assertions are merely documenting the expected single-process signal exit code behavior rather than guarding against this specific regression.

### DIFF
--- a/src/collector.cc
+++ b/src/collector.cc
@@ -18,8 +18,8 @@ class Collector : public ICollector,
 {
 public:
 	Collector(IFileParser &fileParser, IEngine &engine, IFilter &filter) :
-			m_fileParser(fileParser), m_engine(engine), m_exitCode(-1), m_filter(
-					filter)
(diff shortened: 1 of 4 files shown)


SimonKagstrom:
Excellent work, and thanks a lot!

Issue #497, opened 2026-06-25
mach engine: getAligned truncates 64-bit breakpoint addresses (UB shift on macOS)


paulofduarte:
On macOS/x86_64, `src/engines/mach-engine.cc` declares `uint32_t getAligned(uint32_t addr)`, but the callers pass 64-bit `unsigned long` addresses. macOS executables load at a base of `0x1_0000_0000`, so a real breakpoint address such as `0x1000009c2` is truncated to its low 32 bits.

In `arch_setupBreakpoint` this makes `offs = addr - getAligned(addr)` ≈ `0x1_0000_0000`, so `shift = 8 * offs` ≈ `0x8_0000_0000` — far wider than the operand. Built with `-fsanitize=undefined` it aborts:

```
mach-engine.cc: runtime error: shift exponent ... is too large for 64-bit type 'unsigned long'
```

Without the sanitizer the shift count is masked by the hardware, so the result happens to be correct, but it is undefined behaviour. The Linux engine already declares the same helper as `unsigned long getAligned(unsigned long)` (`src/engines/ptrace.cc`).

Fix: widen the macOS `getAligned` to `unsigned long` to match.

PR #498, opened 2026-06-25
Issue #497: widen getAligned to 64-bit in the mach engine

paulofduarte:
Fixes the 64-bit address truncation in #497.

`getAligned` in the mach engine took and returned `uint32_t`, truncating breakpoint addresses above the 4 GB macOS image base and producing an out-of-range shift in `arch_setupBreakpoint` (undefined behaviour; aborts under `-fsanitize=undefined`). Widen it to `unsigned long`, matching the `unsigned long` callers and the Linux ptrace engine's identical helper.

Non-sanitized builds are unaffected — the oversized shift is masked by the hardware shift count — so this only removes the UB and the UBSan abort.

### DIFF
--- a/src/engines/mach-engine.cc
+++ b/src/engines/mach-engine.cc
@@ -69,8 +69,8 @@ constexpr auto regs_flavor = ARM_THREAD_STATE64;
 #endif
 
 
-constexpr uint32_t
-getAligned(uint32_t addr)
+constexpr unsigned long
+getAligned(unsigned long addr)
 {
     return (addr / sizeof(uint32_t)) * sizeof(uint32_t);
 }


SimonKagstrom:
Oh my! Thanks for fixing and investigating!

paulofduarte:
Hi @SimonKagstrom , my pleasure. It took me some time to nail what was happening.

I noticed that there was a failure on the github actions building OSX executable, but it's not related to the change. It seems that just brew installing openssl@1.1 is not available anymore. The earliest version available is openssl@3.0 .

SimonKagstrom:
Thanks for spending the time!

Yes, I also saw that, but as you said it's completely unrelated to your change. There aren't too many builds on kcov these days, so e.g., homebrew package changes can cause issues every now and then...

PR #499, opened 2026-08-06
bash: do not start a here document on a `<<` inside a string

### LINKED ISSUE #472: `echo "<< EOF"` causes rest of file to be ignored
Yes I know it's a bit strange, but there's a good reason for a line like that:  When run in debug mode, my script will output the command, args, and heredoc input to stdout, instead of running the command.  The user can then copy/paste the command to run it independently.

However, when kcov processes my script it stops at the line in the title.

This can be duplicated with the following toy shell script:
```
#!/bin/bash

echo "Yes"
if true; then
	echo 1
fi
echo "<< EOF"
echo "What about this line?"
echo "Nope"
```
The last two lines will run and output when executing the shell script, when run inside of kcov or outside.  However, the kcov results will not show the last two lines as green or red.

kcov_bug

When I comment out the line it works just fine.

I assume something in kcov is detecting heredocs and ignoring them, but that detection is overzealous and the line results in a false positive.

Eljees:
Fixes #472.

## The problem

`BashEngine::parseFileFull` looks for the here-document operator with a plain substring search:

```cpp
size_t heredocStart = s.find("<<");
```

so a `<<` that is part of a string literal is taken for a redirection. In the script from #472:

```bash
echo "<< EOF"
echo "What about this line?"
```

kcov takes `EOF` for the delimiter, switches to the `heredoc` state and skips **every remaining line of the file** looking for a line equal to `EOF` — the rest of the script is reported as non-code.

The same loop already handles quoting when it strips comments a few lines above:

```cpp
else if (comment >= 1 && (s.find("\"") < comment || s.find("'") < comment))
{
    // Do nothing
}
```

so the parser already knows that `#` inside a string is not a comment — it just does not apply the same reasoning to `<<`.

## The fix

A small static helper, `findUnquotedHeredocStart()`, walks the line once tracking single/double quote state (and a backslash escape, which does not apply inside single quotes) and returns the first `<<` that is not inside a string. Everything after that point — the `-` for tab suppression, the trimming, the un-quoting of the delimiter and the `<<<` here-string check — is unchanged.

The change is conservative in one direction only: it can make kcov treat *fewer* lines as the start of a here document, never more.

## Tests

New fixture `tests/bash/heredoc-in-string.sh` and `bash_heredoc_within_string` in `tests/tools/test_bash.py`. Line 22 (`echo "What about this line?"`) has no coverage before the change and is covered after it. The fixture deliberately puts the real here documents *before* the quoted `<<`, so that the terminating `EOF` cannot accidentally end the bogus here document and hide the bug.

The test also pins the behaviour that must not change:

* `cat <<EOF … EOF` — only the first line is code;
* `cat <<-"MARKER" … MARKER` — tab suppression and a quoted delimiter;
* `cat <<<"a here string"` — a here string is not a here document.

The whole `test_bash` suite (31 tests) passes.

## Note on #494

While looking at this I checked #494 as well, and it is a different bug despite the title. The reporter's line (`xiaoshanyan9/git-log-tree@v1.2.1`, lines 133-135) is

```bash
unset "count_sorted_colors[
  color_counts[color]++ << GIT_LOG_TREE_MAX_INT_BITS / 2 | color
]"
```

The middle line, taken on its own, is indistinguishable from `cmd << WORD`: the delimiter is a valid word, there is no `$((`, no `))` and no leading `let `. What actually goes wrong is that the first line opens a double quote and never closes it, and the multi-line-string state in `parseFileFull` only triggers on the `="` / `= "` shapes. So the answer to "I'm not quite sure how to discern the two cases" is that in this particular report you do not have to — the line is inside a string literal.

Worth noting that a rule like "the delimiter must not start with a digit" would be wrong: `x=1<<2` really is a here document for bash itself:

```
warning: here-document at line 3 delimited by end-of-file (wanted `3')
```

I have not touched #494 here; multi-line quote tracking is a much larger change and deserves its own PR.

*AI-assisted: I used Claude to help read the bash engine's state machine and to draft the helper, the fixture and this description. I reviewed every line, ran the new test against unpatched `master` first to confirm line 22 is uncovered without the fix, and ran the full `test_bash` suite locally; the analysis and the runs are mine.*

### DIFF
--- a/src/engines/bash-engine.cc
+++ b/src/engines/bash-engine.cc
@@ -640,15 +640,56 @@ private:
 		}
 	}
 
+	size_t findHeredocStart(const std::string &s, char &quoteChar)
+	{
+		bool escaped = false;
+
+		for (size_t i = 0; i < s.size(); i++)
+		{
+			const char c = s[i];
+
+			if (escaped)
+			{
+				escaped = false;
+				continue;
+			}
+
+			if (c == '\\' && quoteChar != '\'')
+			{
+				escaped = true;
+				continue;
+			}
+
+			if (quoteChar != '\0')
+			{
+				if (c == quoteChar)
+					quoteChar = '\0';
+				continue;
+			}
+
+			if (c == '"' || c == '\'')
+			{
+				quoteChar = c;
+				continue;
+			}
+
+			if (c == '<' && i + 1 < s.size() && s[i + 1] == '<')
+				return i;
+		}
+
+		return std::string::npos;
+	}
+
 	void parseFileFull(const std::string &filename, const std::vector<std::string> &lines, uint32_t crc)
 	{
 		unsigned int lineNo = 0;
 		enum
 		{
-			none, backslash, quote, heredoc
+			none, backslash, heredoc
 		} state = none;
 		bool caseActive = false;
 		bool arithmeticActive = false;
+		char quoteChar = '\0';
 		std::string heredocMarker;
 
 		for (std::vector<std::string>::const_iterator it = lines.begin(); it != lines.end(); ++it)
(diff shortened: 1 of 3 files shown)


SimonKagstrom:
Excellent! Huge kudos for adding the test as well.

The CI failure is unrelated to your PR obviously.

